PR #25, opened 2015-10-07
Bug 1212189 - restore the "sort=oldest" option to the API.

rfk:
This was removed in the sync1.1 -> sync1.5 transition because clients
weren't using it, but we have new clients and they want to use it.

https://bugzilla.mozilla.org/show_bug.cgi?id=1212189


rfk:
(the travis failure here seems unrelated, and only triggered in the postgresql backend)

PR #26, opened 2016-02-24
add fxa uid and device_id to metrics

dannycoates:
This change adds `fxa_uid` and `device_id` from the token to metrics log lines


rfk:
LGTM

PR #28, opened 2016-03-02
Bug 1250747 - make per-batch item limit configurable

rfk:
https://bugzilla.mozilla.org/show_bug.cgi?id=1250747


dannycoates:
LGTM

PR #29, opened 2016-03-09
Bug 1062120 - Fail the whole POST request if we get a DB error

rfk:
https://bugzilla.mozilla.org/show_bug.cgi?id=1062120


rfk:
@Natim it's unrelated to batch-upload stuff, but any chance your could have an r? of this one while you're kind of in a syncstorage headspace?

PR #31, opened 2016-03-28
Add more time.sleep() to functional tests

mostlygeek:
- testing golang implementation, get timestamps that are the
  same on really fast IO systems (SSD)
- Add extra time.sleep() ensures we get unique timestamps

@rfk r?

### DIFF
--- a/syncstorage/tests/functional/test_storage.py
+++ b/syncstorage/tests/functional/test_storage.py
@@ -1298,6 +1298,7 @@ class TestStorage(StorageFunctionalTestCase):
                 self.assertEquals(sorted(int(item['id']) for item in items),
                                   range(start + 1, NUM_ITEMS))
 
+
 class TestStorageMemcached(TestStorage):
     """Storage testcases run against the memcached backend, if available."""
 
--- a/syncstorage/tests/functional/test_storage.py
+++ b/syncstorage/tests/functional/test_storage.py
@@ -142,6 +142,9 @@ class TestStorage(StorageFunctionalTestCase):
         res = self.app.put_json(self.root + '/storage/col2/128', bso)
         ts1 = float(res.headers["X-Last-Modified"])
 
+        # delay to ensure unique timestamp
+        time.sleep(0.01)
+
         bso = {'id': '129', 'payload': 'x'}
         res = self.app.put_json(self.root + '/storage/col2/129', bso)
         ts2 = float(res.headers["X-Last-Modified"])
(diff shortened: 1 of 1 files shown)


rfk:
Thanks @mostlygeek.  I wish this weren't necessary, so I'm going to take a few moments to poke at it later in the week and see if there's something better we can do.

Also, travis is failing due to style errors.

rfk:
@mostlygeek on reflection, I wonder if the reason the python code doesn't need these, is that it uses this handy decorator to do the sleeping automatically in the server itself: https://github.com/mozilla-services/server-syncstorage/blob/master/syncstorage/views/decorators.py#L68

mostlygeek:
@rfk Some of the `time.sleep`, like [this one](https://github.com/mozilla-services/server-syncstorage/blob/af077fe6381612cbe7d3760742f6b288e0228a91/syncstorage/tests/functional/test_storage.py#L278) that tests sort order seem to be required. 

On my macbook pro (late 2015) it's fast enough that those requests get completed in the same 10th of a millisecond.

syncstorage/tests/functional/test_storage.py
```diff
@@ -142,6 +142,9 @@ def test_get_collection(self):
         res = self.app.put_json(self.root + '/storage/col2/128', bso)
         ts1 = float(res.headers["X-Last-Modified"])
 
+        # delay to ensure unique timestamp
+        time.sleep(0.01)
+
```
rfk: So in this case, I believe that a server generating the same timestamp for these requests is in violation of the protocol spec.  We promise the client that sequential write requests will see a monotonically-increasing last-modified timestamp:

  http://docs.services.mozilla.com/storage/apis-1.5.html#concurrency-and-conflict-management

(Yeah, this is non-trivial from the server's point of view, but it's even *more* non-trivial to ask the client to consistently do the right thing here)

syncstorage/tests/functional/test_storage.py
```diff
@@ -272,6 +275,7 @@ def test_get_collection(self):
         for index, sortindex in (('0', -1), ('1', 34), ('2', 12)):
             bso = {'id': index, 'payload': 'x', 'sortindex': sortindex}
             self.app.post_json(self.root + '/storage/col2', [bso])
+            time.sleep(0.01)
```
rfk: Likewise, each independent request here should get an increased timestamp without the client having to worry about it.

mostlygeek:
@rfk I was having #showerthoughts about this and agree. The monotonically increasing timestamp should be enforced on the server side. Closing this PR and adding in some more "tweaks" on the server side.

PR #32, opened 2016-03-31
Largely functional DB query interface

rtilder:
memcache updates need to be flushed out and invalid token responses
need to be sorted out; they're breaking the functional tests for some
odd reason I haven't been able to fathom yet.

### DIFF
--- a/syncstorage/storage/__init__.py
+++ b/syncstorage/storage/__init__.py
@@ -47,8 +47,8 @@ class InvalidOffsetError(StorageError, ValueError):
     pass
 
 
-class InvalidTransaction(StorageError, ValueError):
-    """Exception raised when a request contains an invalid transaction
+class InvalidBatch(StorageError, ValueError):
+    """Exception raised when a request contains an invalid batch
        identifier"""
     pass
 
@@ -314,43 +314,57 @@ class SyncStorage(object):
         """
 
     @abc.abstractmethod
-    def create_transaction(self, userid, collection):
-        """Creates a transaction for multi-POST batch uploads.
+    def create_batch(self, userid, collection):
+        """Creates a batch for multi-POST batch uploads.
 
         Args:
             userid: integer identifying the user in the storage.
             collection: name of the collection.
 
         Returns:
-            The new transaction's ID as a BigInteger timestamp.
+            The new batch's ID as a BigInteger timestamp.
         """
 
     @abc.abstractmethod
-    def append_items_to_transaction(self, batchid, userid, collection, items):
-        """Creates or updates multiple items from a multi-POST batch to a
-        transaction.
+    def valid_batch(self, userid, collection, batchid):
+        """Verifies that a batch ID exists for a given user's collection.
 
         Args:
-            batchid: big integer batch identifier for this transaction
             userid: integer identifying the user in the storage.
             collection: name of the collection.
+            batchid: big integer batch identifier for this batch
+
+        Returns:
+            A value if the batch exists and is still valid and None if the
+            batch ID doesn't exist or is expired or closed.
+        """
+
+    @abc.abstractmethod
+    def append_items_to_batch(self, userid, collection, batchid, items):
+        """Creates or updates multiple items from a multi-POST batch to a
+        batch.
+
+        Args:
+            userid: integer identifying the user in the storage.
+            collection: name of the collection.
+            batchid: big integer batch identifier for this batch
             items: a list of dicts giving data for each item.
 
         Returns:
-            The new last-modified timestamp for the transaction.
+            The new last-modified timestamp for the batch.
 
         Raises:
             ConflictError: the operation conflicted with a concurrent write.
         """
 
     @abc.abstractmethod
-    def commit_transaction(self, batchid, userid, collection):
-        """Commits the pending transaction.
+    def apply_batch(self, userid, collection, batchid):
+        """Applies the pending batch.
 
         Args:
-            batchid: big integer batch identifier for this transaction
             userid: integer identifying the user in the storage.
             collection: name of the collection.
+            batchid: big integer batch identifier for this batch
 
         Returns:
             The new last-modified timestamp for the collection.
@@ -361,13 +375,13 @@ class SyncStorage(object):
         """
 
     @abc.abstractmethod
-    def close_transaction(self, batchid, userid, collection):
-        """Close a specific transaction.
+    def close_batch(self, userid, collection, batchid):
+        """Close a specific batch.
 
         Args:
-            batchid: big integer batch identifier for this transaction
             userid: integer identifying the user in the storage.
             collection: name of the collection.
+            batchid: big integer batch identifier for this batch
         """
 
     #
--- a/syncstorage/storage/__init__.py
+++ b/syncstorage/storage/__init__.py
@@ -47,6 +47,12 @@ class InvalidOffsetError(StorageError, ValueError):
     pass
 
 
+class InvalidTransaction(StorageError, ValueError):
+    """Exception raised when a request contains an invalid transaction
+       identifier"""
+    pass
+
+
 class SyncStorage(object):
     """Abstract Base Class for storage backends.
 
@@ -305,6 +311,63 @@ class SyncStorage(object):
         Raises:
             ConflictError: the operation conflicted with a concurrent write.
             CollectionNotFoundError: the user has no such collection.
+        """
+
+    @abc.abstractmethod
+    def create_transaction(self, userid, collection):
+        """Creates a transaction for multi-POST batch uploads.
+
+        Args:
+            userid: integer identifying the user in the storage.
+            collection: name of the collection.
+
+        Returns:
+            The new transaction's ID as a BigInteger timestamp.
+        """
+
+    @abc.abstractmethod
+    def append_items_to_transaction(self, batchid, userid, collection, items):
+        """Creates or updates multiple items from a multi-POST batch to a
+        transaction.
+
+        Args:
+            batchid: big integer batch identifier for this transaction
+            userid: integer identifying the user in the storage.
+            collection: name of the collection.
+            items: a list of dicts giving data for each item.
+
+        Returns:
+            The new last-modified timestamp for the transaction.
+
+        Raises:
+            ConflictError: the operation conflicted with a concurrent write.
+        """
+
+    @abc.abstractmethod
+    def commit_transaction(self, batchid, userid, collection):
+        """Commits the pending transaction.
+
+        Args:
+            batchid: big integer batch identifier for this transaction
+            userid: integer identifying the user in the storage.
+            collection: name of the collection.
+
+        Returns:
+            The new last-modified timestamp for the collection.
+
+
+        Raises:
+            ConflictError: the operation conflicted with a concurrent write.
+        """
+
+    @abc.abstractmethod
+    def close_transaction(self, batchid, userid, collection):
+        """Close a specific transaction.
+
+        Args:
+            batchid: big integer batch identifier for this transaction
+            userid: integer identifying the user in the storage.
+            collection: name of the collection.
         """
 
     #
--- a/syncstorage/storage/memcached.py
+++ b/syncstorage/storage/memcached.py
@@ -58,7 +58,8 @@ from syncstorage.storage import (SyncStorage,
                                  ConflictError,
                                  CollectionNotFoundError,
                                  ItemNotFoundError,
-                                 InvalidOffsetError)
+                                 InvalidOffsetError,
+                                 InvalidBatch)
 
 from pyramid.settings import aslist
 
@@ -106,7 +107,7 @@ class MemcachedStorage(SyncStorage):
     The SyncStorage implementation wraps another storage backend to provide
     a caching layer.  You may specify the following arguments:
 
-        * storage:  the underlying SyncStorage object that is to be wrapped.a
+        * storage:  the underlying SyncStorage object that is to be wrapped.
         * cache_servers:  a list of memcached server URLs.
         * cached_collections:  a list of names of collections that should
                                be duplicated into memcache for fast access.
@@ -351,6 +352,35 @@ class MemcachedStorage(SyncStorage):
             ts = colmgr.del_items(userid, items)
             update(ts, ts)
             return ts
+
+    def create_batch(self, userid, collection):
+        """Creates batch for a give user's collection."""
+        colmgr = self._get_collection_manager(collection)
+        return colmgr.create_batch(userid)
+
+    def valid_batch(self, userid, collection, batchid):
+        """Verifies that a batch ID is valid"""
+        colmgr = self._get_collection_manager(collection)
+        return colmgr.valid_batch(userid, batchid)
+
+    def append_items_to_batch(self, userid, collection, batchid, items):
+        """Appends items to the pending batch."""
+        colmgr = self._get_collection_manager(collection)
+        ts = colmgr.append_items_to_batch(userid, batchid, items)
+        return ts
+
+    def apply_batch(self, userid, collection, batchid):
+        """Applies the batch"""
+        colmgr = self._get_collection_manager(collection)
+        # prevts = colmgr.get_timestamp(userid)
+        with self._mark_collection_dirty(userid, collection):
+            ts = colmgr.apply_batch(userid, batchid)
+            # Lazy update should occur on following get_item/s
+        return ts
+
+    def close_batch(self, userid, collection, batchid):
+        colmgr = self._get_collection_manager(collection)
+        return colmgr.close_batch(userid, batchid)
 
     #
     # Items APIs
@@ -609,6 +639,27 @@ class UncachedManager(object):
     def del_item(self, userid, item):
         storage = self.owner.storage
         return storage.delete_item(userid, self.collection, item)
+
+    def create_batch(self, userid):
+        storage = self.owner.storage
+        return storage.create_batch(userid, self.collection)
+
+    def valid_batch(self, userid, batchid):
+        storage = self.owner.storage
+        return storage.valid_batch(userid, self.collection, batchid)
+
+    def append_items_to_batch(self, userid, batchid, items):
+        storage = self.owner.storage
+        return storage.append_items_to_batch(userid, self.collection, batchid,
+                                             items)
+
+    def apply_batch(self, userid, batchid):
+        storage = self.owner.storage
+        return storage.apply_batch(userid, self.collection, batchid)
+
+    def close_batch(self, userid, batchid):
+        storage = self.owner.storage
+        return storage.close_batch(userid, self.collection, batchid)
 
 
 class _CachedManagerBase(object):
@@ -880,6 +931,85 @@ class CacheOnlyManager(_CachedManagerBase):
             raise ItemNotFoundError
         return modified
 
+    def get_batches_key(self, userid):
+        return _key(userid, "c", self.collection, "batches")
+
+    def get_cached_batches(self, userid):
+        return self.cache.gets(self.get_batches_key(userid))
+
+    def create_batch(self, userid):
+        bdata, bcasid = self.get_cached_batches(userid)
+        batch = get_timestamp()
+        batchid = int(batch * 1000)
+        if not bdata:
+            bdata = {}
+        if batchid in bdata:
+            raise ConflictError
+        bdata[batchid] = {"user": userid,
+                          "modified": batch,
+                          # FIXME Rough guesstimate of the maximum
+                          #       reasonable life span of a batch
+                          "expires": int(batch) + 2 * 3600,
+                          "items": []}
+        key = self.get_batches_key(userid)
+        if not self.cache.cas(key, bdata, bcasid):
+            raise ConflictError
+        return batchid
+
+    def valid_batch(self, userid, batch):
+        ts = get_timestamp()
+        batchid = str(batch)
+        bdata, bcasid = self.get_cached_batches(userid)
+
+        if batchid not in bdata or bdata[batchid]["expires"] < ts:
+            self.close_batch(userid, batchid)
+            return False
+        return True
+
+    def append_items_to_batch(self, userid, batch, items):
+        modified = get_timestamp()
+        batchid = str(batch)
+        bdata, bcasid = self.get_cached_batches(userid)
+        # Invalid, closed, or expired batch
+        if (not bdata or
+                batchid not in bdata or
+                bdata[batchid]["expires"] <= int(modified)):
+            raise InvalidBatch(batch, modified, bdata)
+
+        bdata[batchid]["items"].extend(items)
+        bdata[batchid]["modified"] = modified
+        key = self.get_batches_key(userid)
+        if not self.cache.cas(key, bdata, bcasid):
+            raise ConflictError
+        return modified
+
+    def apply_batch(self, userid, batch):
+        modified = get_timestamp()
+        batchid = str(batch)
+        bdata, bcasid = self.get_cached_batches(userid)
+        # Invalid, closed, or expired batch
+        if (not bdata or
+                batchid not in bdata or
+                bdata[batchid]["expires"] <= int(modified)):
+            raise InvalidBatch(batch, modified, bdata)
+
+        data, casid = self.get_cached_data(userid)
+        self._set_items(userid, bdata[batchid]["items"], modified, data, casid)
+        return modified
+
+    def close_batch(self, userid, batch):
+        batchid = str(batch)
+        bdata, bcasid = self.get_cached_batches(userid)
+        key = self.get_batches_key(userid)
+
+        if batchid in bdata:
+            try:
+                del bdata[batchid]
+            except KeyError:
+                return
+        if not self.cache.cas(key, bdata, bcasid):
+            raise ConflictError
+
 
 class CachedManager(_CachedManagerBase):
     """Object for managing storage of a collection in both cache and store.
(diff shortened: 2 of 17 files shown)


syncstorage/storage/__init__.py
```diff
@@ -307,6 +313,63 @@ def delete_items(self, userid, collection, items):
...
+            The new transaction's ID as a BigInteger timestamp.
+        """
+
+    @abc.abstractmethod
+    def append_items_to_transaction(self, batchid, userid, collection, items):
+        """Creates or updates multiple items from a multi-POST batch to a
+        transaction.
+
+        Args:
+            batchid: big integer batch identifier for this transaction
```
rfk: The naming confusion between "batchid" and "transaction" really stands out in the code here; if the name of the thing is to be "transaction" then lets make this "trnid" or "transactionid" or something all throughout the code and the API.

syncstorage/storage/__init__.py
```diff
@@ -307,6 +313,63 @@ def delete_items(self, userid, collection, items):
...
+            userid: integer identifying the user in the storage.
+            collection: name of the collection.
+            items: a list of dicts giving data for each item.
+
+        Returns:
+            The new last-modified timestamp for the transaction.
```
rfk: It's not obvious to me why this timestamp would be useful, and if we're not going to use it then perhaps we should just return `None`.

syncstorage/storage/__init__.py
```diff
@@ -307,6 +313,63 @@ def delete_items(self, userid, collection, items):
...
+
+        Raises:
+            ConflictError: the operation conflicted with a concurrent write.
+        """
+
+    @abc.abstractmethod
+    def commit_transaction(self, batchid, userid, collection):
```
rfk: For consistency with the other methods on this class, perhaps the order of arguments here should be `userid`, `collection`, `batchid`.

syncstorage/views/__init__.py
```diff
@@ -295,6 +315,9 @@ def post_collection(request):
     bsos = request.validated["bsos"]
     invalid_bsos = request.validated["invalid_bsos"]
 
+    if request.validated["batch"] or request.validated["commit"]:
+        return post_collection_transaction(request)
+
```
rfk: It may be worth pulling the original non-transactional implementation out into a separate `post_collection_single` function or similar, to make it more obvious that we're choosing between two independent implementations.

syncstorage/views/__init__.py
```diff
@@ -319,6 +342,84 @@ def post_collection(request):
...
+    batch = request.validated["batch"]
+    commit = request.validated["commit"]
+
+    # The "batch" key is set only on a multi-POST batch request prior to a
+    # commit.  The "modified" key is only set upon a successful commit.
+    # The two flags are mutually exclusive.
+    # Any failures at all mean cancelling the transaction completely.
+    res = {'success': [], 'failed': {}}
+
+    # If there are any parsing failures, we won't even start a transaction.
```
rfk: We should check this behaviour with @rnewman, since what you describe differs from behaviour in the non-transactional case.  In that case the valid bsos are written and the invalid ones are rejected, and I can see an argument that we should do the same here, report the failures to the client, and let it decide what to do about the problem.

syncstorage/views/__init__.py
```diff
@@ -319,6 +342,84 @@ def post_collection(request):
...
+    try:
+        if batch is True:
+            try:
+                batch = storage.create_transaction(userid, collection)
+            except ConflictError, e:
+                # ConflictError here means a client is spamming requests,
+                # I think.
+                logger.error('Collision in transaction creation!')
+                logger.error(e)
+                raise HTTPServiceUnavailable('Too many transactions opened')
```
rfk: I think we should raise the original ConflictError here, and it'll be converted into an appropriate response by the middleware layer (I believe it's a 503)

syncstorage/views/__init__.py
```diff
@@ -319,6 +342,84 @@ def post_collection(request):
...
+                         '("{0}")'.format(batch))
+            logger.error(str(e))
+            for bso in bsos:
+                res["failed"][bso["id"]] = "db error"
+            raise
+
+        if commit:
+            try:
+                storage.commit_transaction(batch, userid, collection)
+                res['modified'] = ts
```
rfk: IIRC, we want all items in the transaction to get the timestamp from the time of commit.  It may be slightly more reliable to use `ts = store.commit_transaction()` here rather than using the timestamp from the previous `append_items_to_transaction`.

syncstorage/views/__init__.py
```diff
@@ -319,6 +342,84 @@ def post_collection(request):
...
+                request.response.headers["X-Last-Modified"] = str(ts)
+            except ConflictError:
+                raise
+            except Exception, e:
+                logger.error("Could not commit transaction")
+                logger.error(e)
+                raise
+            finally:
+                for bso in bsos:
+                    res["failed"][bso["id"]] = "db error: commit"
```
rfk: Will this `finally` block trigger even if there's no exception, reporting all bsos as failed?

syncstorage/views/__init__.py
```diff
@@ -319,6 +342,84 @@ def post_collection(request):
...
+        else:
+            res["batch"] = batch
+        res["success"].extend([bso["id"] for bso in bsos])
+    except ConflictError:
+        raise
+    finally:
+        if batch and batch is not True:
+            storage.close_transaction(batch, userid, collection)
```
rfk: Likewise, won't this result in `close_transaction` being called on every request except the initial one?

syncstorage/views/validators.py
```diff
@@ -146,6 +146,31 @@ def extract_query_params(request):
...
+    is to be committed and deleted.
+    """
+    request.validated["batch"] = False
+    batch_id = request.GET.get("batch")
+    if batch_id is not None:
+        if batch_id == "true":
+            batch_id = True
+        request.validated["batch"] = batch_id
+    elif batch_id is None and "batch" in request.GET:
+        request.validated["batch"] = True
```
rfk: Is it actually possible for `"batch"` to wind up in `request.GET` with a value of `None`?

syncstorage/views/validators.py
```diff
@@ -146,6 +146,31 @@ def extract_query_params(request):
...
+
+    request.validated["commit"] = False
+    commit = request.GET.get("commit")
+    if commit is not None and commit == "true":
+        request.validated["commit"] = True
```
rfk: If there's a value other than `"true"` we should probably return a 400 error.

syncstorage/views/__init__.py
rnewman: This is an aspect of the Sync protocol that I kinda hate: Sync essentially operates like `INSERT OR IGNORE`, including for malformed input, which is so rarely what one wants!

Fortunately, in the case of batched uploads, the client can see that some of the uploaded BSOs were rejected, and can choose to not commit the entire transaction. 

I think it's fine to keep the existing behavior (queue the ones that were fine) for that reason; if anything, the sane thing to do the reverse: to reject part-wrong non-batch `POST`s and part-accept batched ones!

Two observations:

- It'd be nice to have a `?discard` operation, just for efficiency's sake.
- Careful clients should always issue an empty `POST` to commit, so that they don't accidentally commit a partial final batch.

syncstorage/tests/functional/test_storage.py
```diff
@@ -1286,6 +1287,49 @@ def test_pagination_with_newer_and_sort_by_oldest(self):
...
+        self.app.post_json(endpoint, bsos)
+
+        bso3 = {'id': 'a', 'payload': 'internal'}
+        bso4 = {'id': 'b', 'payload': 'pancreas'}
+        resp = self.app.post_json(endpoint + '?batch=true', [bso3, bso4])
+        batch = resp.json["batch"]
+        # This is sooooo going to break when it's run at just the wrong time
+        # and will result in a seriously confused person at the keyboard.
+        guess = float(resp.headers["X-Weave-Timestamp"])
+        self.assertEqual(batch / 1000, math.floor(guess))
```
rfk: Since the value of batch is defined to be opaque, we probably don't need to check that it has any particular value, but rather just check that it's an appropriate-looking string.

syncstorage/tests/functional/test_storage.py
```diff
@@ -1286,6 +1287,49 @@ def test_pagination_with_newer_and_sort_by_oldest(self):
...
+
+        # make sure we don't have the pending commit
+        resp = self.app.get(endpoint)
+        res = resp.json
+        res.sort()
+        self.assertEquals(res, ['12', '13'])
+        self.assertEquals(int(resp.headers['X-Weave-Records']), 2)
```
rfk: It'd also be worth checking that the timestamp on the collection hasn't been modified yet.

syncstorage/tests/functional/test_storage.py
```diff
@@ -1286,6 +1287,49 @@ def test_pagination_with_newer_and_sort_by_oldest(self):
...
+        res = resp.json
+        res.sort()
+        self.assertEquals(res, ['12', '13', 'a', 'b', 'c'])
+        self.assertEquals(int(resp.headers['X-Weave-Records']), 5)
+        resp = self.app.get(endpoint + '/13')
+        self.assertEquals(resp.json["payload"], 'portnoy')
+
+        # invalid transaction ID
+        bso7 = {'id': 'd', 'payload': 'burrito'}
+        resp = self.app.post_json(endpoint + '?batch=sammich', [bso7])
```
rfk: Please break these error-case tests out into one or more separate test methods.

syncstorage/tests/functional/test_storage.py
```diff
@@ -1286,6 +1287,49 @@ def test_pagination_with_newer_and_sort_by_oldest(self):
...
+                          float(resp.headers['X-Weave-Timestamp']))
+
+        # make sure the changes apply
+        resp = self.app.get(endpoint)
+        res = resp.json
+        res.sort()
+        self.assertEquals(res, ['12', '13', 'a', 'b', 'c'])
+        self.assertEquals(int(resp.headers['X-Weave-Records']), 5)
+        resp = self.app.get(endpoint + '/13')
+        self.assertEquals(resp.json["payload"], 'portnoy')
```
rfk: There are probably additional consistency checks worth applying here, e.g. checking that all the items in the batch got assigned the same timestamp, and that it matches the timestamp reported in the `commit` response above.

rfk: In fact, @rnewman, I encourage you to think of any specific edge-case-style interactions that might affect this API, and we can add specific tests for them.

syncstorage/storage/sql/__init__.py
```diff
@@ -456,6 +456,54 @@ def set_items(self, session, userid, collection, items):
...
+    def append_items_to_transaction(self, session, batchid, userid,
+                                    collection, items):
+        """Inserts items into batch_upload_items"""
+        rows = []
+        for data in items:
+            id_ = data["id"]
+            row = self._prepare_bui_row(session, batchid, id_, data)
+            rows.append(row)
+        defaults = {"modified": ts2bigint(session.timestamp)}
+        session.insert_or_update("batch_upload_items", rows, defaults)
```
rfk: Per https://bugzilla.mozilla.org/show_bug.cgi?id=1250189#c16, I suspect we want all items in the transaction to receive the same modification time, and for that time to be the time of the final commit.

syncstorage/storage/sql/dbconnect.py
```diff
@@ -112,19 +112,33 @@ def __new__(cls, *args, **kwds):
...
-               autoincrement=False),
-        Column("collection", Integer, primary_key=True, nullable=False,
-               autoincrement=False),
-        Column("id", String(64), primary_key=True, autoincrement=False),
         Column("sortindex", Integer),
         Column("modified", BigInteger, nullable=False),
+        # MySQL no likey "DEFAULT ''".  Results in an error:
+        #     ERROR 1101 (42000): BLOB, TEXT, GEOMETRY or JSON column 'payload'
+        #                         can't have a default value
+        # Column("payload", PAYLOAD_TYPE, nullable=False),
```
rfk: Ouch, not sure why we haven't hit this before :-/.  I guess we may as well fix it as part of this work, unless it turns out to be a rabbit-hole.

syncstorage/storage/sql/dbconnect.py
```diff
@@ -144,21 +159,85 @@ def _get_bso_columns(table_name):
...
+        Column("item", String(64), primary_key=True, nullable=False,
+               autoincrement=False)
+    )
+    common = _common_columns(MAX_TTL)
+    indexes = (
+        # Declare indexes.
+        # We need to include the tablename in the index name due to sharding,
+        # because index names in sqlite are global, not per-table.
+        # Index on "ttl" for easy pruning of expired items.
+        Index("%s_ttl_idx" % (table_name,), "ttl"),
```
rfk: I'm not convinced we'll ever want to prune expired items from this table, but rather just prune everything that's been committed.  In fact, do we need anything other than the primary key index on this table?

syncstorage/storage/sql/dbconnect.py
```diff
@@ -144,21 +159,85 @@ def _get_bso_columns(table_name):
 
 bso = Table("bso", metadata, *_get_bso_columns("bso"))
 
+# Table mapping (user_id, collection_id) => batch IDs
+
+batch_uploads = Table(
+    "batch_uploads",
+    metadata,
+    Column("batch", BigInteger, primary_key=True, nullable=False),
```
rfk: If we're using just a timestamp as the batch id, we'll almost certainly generate spurious collisions between multiple users trying to create a transaction at the same time.  We could make this a compound primary key on `(batch, userid)`, or make `batch` a string and include more information that way.  @rtilder thoughts?

syncstorage/storage/sql/dbconnect.py
```diff
@@ -144,21 +159,85 @@ def _get_bso_columns(table_name):
...
+    Column("userid", Integer, nullable=False, autoincrement=False),
+    Column("collection", Integer, nullable=False, autoincrement=False),
```
rfk: You shouldn't need `autoincrement=False` on these non-primary-key columns.

syncstorage/storage/sql/dbconnect.py
```diff
@@ -371,6 +454,11 @@ def get_query(self, name, params):
                 qvars["bso"] = params["bso"]
             else:
                 qvars["bso"] = self.get_bso_table(params["userid"])
+        if "%(batch_upload_items)s" in query:
```
rfk: nit: the inconsistency between abbreviated `%(bso)s` and long-form `%(batch_upload_items)s` is a little strange.

syncstorage/storage/sql/dbconnect.py
```diff
@@ -387,6 +475,12 @@ def get_bso_table(self, userid):
             return bso
         return get_bso_table(userid % self.shardsize)
 
+    def get_batch_item_table(self, userid):
+        """Get the batch_upload_items table object for the given userid."""
+        if not self.shard or userid is None:
+            return bui
+        return get_bso_table(userid % self.shardsize)
```
rfk: s/get_bso_table/get_batch_item_table/g

syncstorage/storage/sql/queries_generic.py
```diff
@@ -12,6 +12,7 @@
 string interpolation variables with special meaning:
 
     * %(bso)s:   insert the name of the user's sharded BSO storage table
+    * %(bui)s:   insert the name of the user's sharded batch_upload_items table
```
rfk: documentation doesn't match code here, code uses `%(batch_upload_items)s`

syncstorage/storage/sql/queries_generic.py
```diff
@@ -92,6 +93,47 @@
...
+                     "   payload, payload_size, ttl) "\
+                     "SELECT "\
+                     "  :userid, :collection, item, sortindex, modified, "\
+                     "  COALESCE(payload, \"\"), COALESCE(payload_size, 0), "\
+                     "  COALESCE(ttl, ttl + :default_ttl) "\
+                     "FROM %(batch_upload_items)s "\
+                     "WHERE batch = :batch "\
+                     "ON DUPLICATE KEY UPDATE "\
+                     "  sortindex = COALESCE(VALUES(%(bso)s.sortindex), " \
+                     "                   %(batch_upload_items)s.sortindex), " \
```
rfk: The order of arguments to `COALESCE` seems off here...IIUC we want to take the value from `batch_upload_items` if it's non-NULL, but fall back to keeping the value from `bso` if it is NULL.  So the arguments should be the other way around.

syncstorage/storage/sql/queries_generic.py
```diff
@@ -92,6 +93,47 @@
...
+COMMIT_TRANSACTION = "INSERT INTO %(bso)s "\
+                     "  (userid, collection, id, sortindex, modified, "\
+                     "   payload, payload_size, ttl) "\
+                     "SELECT "\
+                     "  :userid, :collection, item, sortindex, modified, "\
+                     "  COALESCE(payload, \"\"), COALESCE(payload_size, 0), "\
+                     "  COALESCE(ttl, ttl + :default_ttl) "\
+                     "FROM %(batch_upload_items)s "\
+                     "WHERE batch = :batch "\
+                     "ON DUPLICATE KEY UPDATE "\
```
rfk: Since `ON DUPLICATE KEY UPDATE` is a MySQL-specific extension, this version of the query should go into `queries_mysql` and a more generic one should be used here.  Is the sqlite version generic enough to work as the default implementation?

syncstorage/storage/sql/queries_generic.py
```diff
@@ -163,3 +205,8 @@ def FIND_ITEMS(bso, params):
 # version using DELETE <blah> LIMIT 1000.
 PURGE_SOME_EXPIRED_ITEMS = "DELETE FROM %(bso)s "\
                            "WHERE ttl < (UNIX_TIMESTAMP() - :grace) "
+
+PURGE_TRANSACTIONS = "DELETE FROM batch_uploads WHERE batch < :batch"
+
+PURGE_TRANSACTION_CONTENTS = "DELETE FROM batch_upload_items WHERE batch < " \
+                             ":batch"
```
rfk: Looks good, we'll also need a script to run periodically that will perform this purge.

syncstorage/storage/sql/queries_generic.py
```diff
@@ -92,6 +93,47 @@
 DELETE_ITEMS = "DELETE FROM %(bso)s WHERE userid=:userid "\
                "AND collection=:collectionid AND id IN %(ids)s"
 
+CREATE_TRANSACTION = "INSERT INTO batch_uploads (batch, userid, collection) "\
+                     "VALUES (:batch, :userid, :collection)"
+
+VALID_TRANSACTION = "SELECT open FROM batch_uploads WHERE batch = :batch " \
+                    "AND userid = :userid AND collection = collection"
```
rfk: This isn't used anywhere yet.

In general, what's the advantage of setting `open=0` as opposed to e.g. just deleting the entry from `batch_uploads` once we're done with it?

syncstorage/storage/sql/__init__.py
```diff
@@ -456,6 +456,54 @@ def set_items(self, session, userid, collection, items):
...
+    @with_session
+    def commit_transaction(self, session, batchid, userid, collection):
+        collectionid = self._get_collection_id(session, collection)
+        params = {
+            "batch": batchid,
+            "userid": userid,
+            "collection": collectionid,
+            "default_ttl": MAX_TTL
+        }
+        session.query("COMMIT_TRANSACTION", params)
```
rfk: I wonder whether it's worth adding an explicit timer around this new and scary query, so we can monitor it more closely than the mysql slow-query log might allow.

syncstorage/storage/sql/dbconnect.py
rtilder: D'oh!  It's supposed to be PRIMARY KEY (batch, userid).  Looks like this plus the autoincrement=False on the collection column are victims are failed copying and pasting.

syncstorage/storage/sql/dbconnect.py
rtilder: Nope.  It's just deliberate mimicry of the bso table indexes.  No real need.

syncstorage/storage/sql/dbconnect.py
rtilder: There's my.ini setting to honor MySQL 4.x semantics that I suspect is more common than not.  I'll find the specifics in my notes and see if I can figure out what TravisCI's default setting is.

syncstorage/views/__init__.py
rtilder: Discard entire group[1] of batches?

1. Should be murder of batches.

syncstorage/storage/sql/dbconnect.py
rtilder: Additional thoughts: changing the type for `batch` means we can use version 4 UUIDs as identifiers directly without dealing with converting to and from integers to store them.  I'm not certain if that would affect the `batch_upload_items` table's purging performance.  I believe it wouldn't but would definitely want to verify experimentally first.

syncstorage/storage/__init__.py
```diff
@@ -47,6 +47,12 @@ class InvalidOffsetError(StorageError, ValueError):
     pass
 
 
+class InvalidBatch(StorageError, ValueError):
```
rfk: naming nit: this should be with `InvalidBatchError` for consistency with its sibling class names.

syncstorage/storage/memcached.py
```diff
@@ -880,6 +931,85 @@ def del_item(self, userid, item):
...
+        batchid = int(batch * 1000)
+        if not bdata:
+            bdata = {}
+        if batchid in bdata:
+            raise ConflictError
+        bdata[batchid] = {"user": userid,
+                          "modified": batch,
+                          # FIXME Rough guesstimate of the maximum
+                          #       reasonable life span of a batch
+                          "expires": int(batch) + 2 * 3600,
```
rfk: We should read this from config, and use it uniformly in both backends

syncstorage/storage/sql/dbconnect.py
```diff
@@ -112,19 +112,39 @@ def __new__(cls, *args, **kwds):
...
+        # into a warning:
+        #
+        # [mysqld]
+        # sql_mode=MYSQL40
+        #
         Column("payload", PAYLOAD_TYPE, nullable=False, server_default=""),
         Column("payload_size", Integer, nullable=False,
                server_default=sqltext("0")),
-        Column("ttl", Integer, server_default=sqltext(str(MAX_TTL))),
+        Column("ttl", Integer, server_default=sqltext(str(ttl)))
```
rfk: All calls to this function pass in `MAX_TTL` as the arguments, we can probably simplify by just continuing to refer to it directly here.

syncstorage/storage/sql/queries_generic.py
```diff
@@ -92,6 +93,46 @@
 DELETE_ITEMS = "DELETE FROM %(bso)s WHERE userid=:userid "\
                "AND collection=:collectionid AND id IN %(ids)s"
 
+CREATE_BATCH = "INSERT INTO batch_uploads (batch, userid, collection) "\
+                     "VALUES (:batch, :userid, :collection)"
+
+VALID_BATCH = "SELECT batch FROM batch_uploads WHERE batch = :batch " \
+                    "AND userid = :userid AND collection = :collection"
+
+APPEND_ITEMS_TO_BATCH = "INSERT INTO %(bui)s "\
```
rfk: This query appears to be unused, I assume because of use of the `insert_or_update` helper.

syncstorage/storage/sql/queries_generic.py
```diff
@@ -92,6 +93,46 @@
...
+                              "  payload_size = VALUES(payload_size) "\
+                              "  ttl = VALUES(ttl)"
+
+APPLY_BATCH = "INSERT INTO %(bso)s "\
+                     "  (userid, collection, id, sortindex, modified, "\
+                     "   payload, payload_size, ttl) "\
+                     "SELECT "\
+                     "  :userid, :collection, id, sortindex, :modified, " \
+                     "  COALESCE(payload, \"\"), COALESCE(payload_size, 0), "\
+                     "  COALESCE(ttl, ttl + :default_ttl) "\
```
rfk: Since the `ttl` column as a default value, I suspect we can just select `ttl` here and let the default do the right thing, rather than trying to `COALESCE` it to a specific non-`NULL` value.

syncstorage/storage/sql/queries_generic.py
```diff
@@ -92,6 +93,46 @@
...
+                     "FROM %(bui)s "\
+                     "WHERE batch = :batch "\
```
rfk: now that we have a composite primary key, we need to add `userid` to the `WHERE` clause here.

syncstorage/tests/functional/test_storage.py
```diff
@@ -1286,6 +1287,97 @@ def test_pagination_with_newer_and_sort_by_oldest(self):
                 self.assertEquals(sorted(int(item['id']) for item in items),
                                   range(start + 1, NUM_ITEMS))
 
+    def assertCloseEnough(self, val1, val2, delta=0.05):
+        if abs(val1 - val2) < delta:
+            return True
+        raise AssertionError("abs(%.2f - %.2f) = %.2f > %.2f"
+                             % (val1, val2, abs(val1 - val2), delta))
```
rfk: nice :-)

rfk:
I left a couple comments but still need to come back for another pass through this.  Concurrently, we should also try to get this branch deployed somwhere that we can take it for a spin in the real world.

rfk:
Closing this and bringing it back as a PR with branch in the mozilla-services repo

PR #33, opened 2016-04-15
Use correct name for PURGE_SOME_EXPIRED_ITEMS in mysql specific queries.

rfk:
@l-hedgehog does this look OK to you?


rfk:
@rtilder could you please r? me on this trivial bugfix, just to be safe?

l-hedgehog:
Hi @rfk, I think this is the right fix. Thanks!

PR #34, opened 2016-05-24
Dont try to set a default on a MySQL TEXT column.

rfk:
This is not supported in newer versions of MySQL.  @rtilder r?


rtilder:
r+

PR #36, opened 2016-06-06
Batch uploads

rfk:
This is #32 re-opened with a branch in this repo


loadtest/stress.py
```diff
@@ -176,9 +176,25 @@ def test_storage_session(self):
             response = self.session.get(url, params=params, **reqkwds)
             self.assertTrue(response.status_code in (200, 404))
 
-        # PUT requests with several WBOs batched together
+        # POST requests with several WBOs batched together
         num_requests = self._pick_weighted_count(post_count_distribution)
-        cols = random.sample(collections, num_requests)
+        # Shallower distribution(for now) of "transactional" style batch
```
rfk: nit: space after "distribution"

loadtest/stress.py
```diff
@@ -176,9 +176,25 @@ def test_storage_session(self):
...
+        # uploads.  Looking for roughly 10% of batch POSTs to be
+        # "transactional".  Increase the frequency of "transactional" requests
+        # by changing the modulo 10 to something like modulo 5(~20%) or
+        # modulo 3(~30%).
+        # transact = int(random.paretovariate(1) * 100) % 10 == 0
+        #
+        # Let's do roughly 50% for fun.
+        transact = random.randint(0, 1)
```
rfk: does this change mean the comment above it about the "modulo blah" is now obsolete?

syncstorage/storage/memcached.py
```diff
@@ -880,6 +931,85 @@ def del_item(self, userid, item):
...
+        batch = get_timestamp()
+        batchid = int(batch * 1000)
+        if not bdata:
+            bdata = {}
+        if batchid in bdata:
+            raise ConflictError
+        bdata[batchid] = {"user": userid,
+                          "modified": batch,
+                          # FIXME Rough guesstimate of the maximum
+                          #       reasonable life span of a batch
```
rfk: We should be able to read this from configuration somewhere; if not, it's probably something we should add to the configuration somewhere.

syncstorage/storage/memcached.py
```diff
@@ -880,6 +931,85 @@ def del_item(self, userid, item):
...
+    def create_batch(self, userid):
+        bdata, bcasid = self.get_cached_batches(userid)
+        batch = get_timestamp()
+        batchid = int(batch * 1000)
+        if not bdata:
+            bdata = {}
+        if batchid in bdata:
+            raise ConflictError
+        bdata[batchid] = {"user": userid,
+                          "modified": batch,
```
rfk: AFAICT we do not use either of the "user" or "modified" keys from this object.  If true we should just drop them and save the overhead.

syncstorage/storage/memcached.py
```diff
@@ -880,6 +931,85 @@ def del_item(self, userid, item):
...
+        batchid = str(batch)
+        bdata, bcasid = self.get_cached_batches(userid)
+        # Invalid, closed, or expired batch
+        if (not bdata or
+                batchid not in bdata or
+                bdata[batchid]["expires"] <= int(modified)):
+            raise InvalidBatch(batch, modified, bdata)
+
+        bdata[batchid]["items"].extend(items)
+        bdata[batchid]["modified"] = modified
```
rfk: As above, I don't think we actually use this "modified" timestamp anywhere?

syncstorage/storage/memcached.py
```diff
@@ -880,6 +931,85 @@ def del_item(self, userid, item):
...
+        modified = get_timestamp()
+        batchid = str(batch)
+        bdata, bcasid = self.get_cached_batches(userid)
+        # Invalid, closed, or expired batch
+        if (not bdata or
+                batchid not in bdata or
+                bdata[batchid]["expires"] <= int(modified)):
+            raise InvalidBatch(batch, modified, bdata)
+
+        bdata[batchid]["items"].extend(items)
```
rfk: We should check that the total size of the batch has not exceeded our limit, otherwise clients will be allowed to just accumulate data in memcached indefinitely.  It might even be worth putting some limit on the number of active batches per userid as an additional safely measure.

syncstorage/storage/sql/__init__.py
```diff
@@ -460,6 +462,72 @@ def set_items(self, session, userid, collection, items):
...
+        for data in items:
+            id_ = data["id"]
+            row = self._prepare_bui_row(session, batchid, id_, data)
+            rows.append(row)
+        defaults = {
+            "modified": ts2bigint(session.timestamp),
+            "payload": "",
+            "payload_size": 0,
+        }
+        session.insert_or_update("batch_upload_items", rows, defaults)
```
rfk: I wish there were an obvious way to check that the size of the batch has not gotten too big...but I can't think of anything better than a "SELECT COUNT()" and I don't think we want to pay the overhead of that.

syncstorage/storage/sql/__init__.py
```diff
@@ -460,6 +462,72 @@ def set_items(self, session, userid, collection, items):
...
+    def valid_batch(self, session, userid, collection, batchid):
+        """Checks to see if the batch ID is valid and still open"""
+        collectionid = self._get_collection_id(session, collection)
+        params = {
+            "batch": batchid,
+            "userid": userid,
+            "collection": collectionid
+        }
+        valid = session.query_scalar("VALID_BATCH", params=params)
+        return valid
```
rfk: Since the batchid is a timestamp, we should be able to answer `false` to `valid_batch` if called with a very old batch id, outside whatever the expiry time is for a batch.

syncstorage/storage/sql/__init__.py
```diff
@@ -593,55 +670,91 @@ def delete_item(self, session, userid, collection, item):
...
-        else:
-            tables = set(self.dbconnector.get_bso_table(i).name
-                         for i in xrange(self.dbconnector.shardsize))
-            assert len(tables) == self.dbconnector.shardsize
-        # Purge each table in turn, summing rowcounts.
-        # We set an upper limit on the number of iterations, to avoid
-        # getting stuck indefinitely on a single table.
-        total_affected = 0
+           The grace period for batch_upload_items is hard coded to 1.5 time
+           the current upper limit(2 hours) for a batch session's lifetime.
```
rfk: nit: space after "limit"?

Does this imply that the batch session lifetime of 2 hours is also hard-coded, implicity at least?  Seems like something we should lift into config if so.

syncstorage/storage/sql/__init__.py
```diff
@@ -593,55 +670,91 @@ def delete_item(self, session, userid, collection, item):
...
+        """
+
+        table_sets = {"bso": {"replace": "bso",
+                              "query": "PURGE_SOME_EXPIRED_ITEMS",
+                              "total_affected": 0},
+                      "batch_upload_items": {"replace": "bui",
+                                             "query": "PURGE_BATCH_CONTENTS",
+                                             "total_affected": 0,
+                                             "grace_period": 3 * 60 * 60}}
```
rfk: We could take the `max()` of this pre-set limit and the grace_period provided by the caller.

syncstorage/storage/sql/__init__.py
```diff
@@ -593,55 +670,91 @@ def delete_item(self, session, userid, collection, item):
...
+                            tabula["replace"]: table,
+                            "grace": tabula.get("grace_period", grace_period),
+                            "maxitems": max_per_loop,
+                        })
+                logger.info("Purged %d expired items from %s",
+                            num_affected, table)
+                tabula["total_affected"] += num_affected
+
+        logger.info("Purged %d expired batches from "
+                    "batch_uploads" % total_batches_purged)
```
rfk: Wow, this purge method wound up getting pretty gnarly.  I wonder if it would be clearer (if slightly more verbose) to make three separate methods `purge_expired_bsos`, `purge_expired_batches` `purge_expired_batch_items` rather than squish them all into a single control-flow here.

syncstorage/storage/sql/__init__.py
```diff
@@ -593,55 +670,91 @@ def delete_item(self, session, userid, collection, item):
...
+                             for i in xrange(self.dbconnector.shardsize))
+                assert len(tables) == self.dbconnector.shardsize
+            # Purge each table in turn, summing rowcounts.
+            # We set an upper limit on the number of iterations, to avoid
+            # getting stuck indefinitely on a single table.
+            for table in sorted(tables):
+                logger.info("Purging expired items from %s", table)
+                num_iters = 1
+                num_affected = 0
+                with self._get_or_create_session() as session:
```
rfk: I wonder if the thing where we're creating a fresh session here, rather than holding an open session for the duration of the entire purge, is contributing to the surprisingly good behaviour of this code under loadtest.  Even if not, +1 to this change!

rfk: (If it *is* helping performance, it may be worth leaving a comment about why it's important to re-create the session on each iteration.  Future me is very likely to forget that this was important.)

rfk:
Aaaand we're done!  Thanks @Natim and @rtilder, I've merged this to master.

mostlygeek:
YAY!

PR #37, opened 2016-06-06
Tweaks to get tests passing on batch uploads PR

rfk:
@rtilder some tweaks to your db queries:

* Move queries with mysql-specific `ON DUPLICATE KEY` logic into `queries_mysql.py`
* Fix syntax error in sqlite `APPLY_BATCH` query due to mis-embedded double quotes
* Try to make some postgres-specific purge queries
* Remove `APPEND_ITEMS_TO_BATCH` which doesn't seem to be used

It doesn't get all the tests passing, but they're passing more than they were for non-mysql backends. @rtilder r?


rfk:
@rtilder I got a bit obsessed with trying to get the tests passing here, so have added a commit implementing the checking of `X-Weave-Records` and friends.  Let's see what travis says about it...

rfk:
w00t, tests passing!

rtilder:
Looks great to me.  I'm all for merging for use on stage.

PR #38, opened 2016-06-07
fix functional tests failing against https endpoints

mostlygeek:
Added `ca-certificates` to the docker build requirements. Now `test_functional` will work against https endpoints. It failed before because openssl and the ca-certificates were not installed. 

@rfk r?


mostlygeek:
Note: tested the container with the ca-certificates and it works testing live sync servers in stage.

rfk:
👍 ; just waiting on travis for the green light

PR #41, opened 2016-09-08
Bug 1300451 - log data about invalid BSOs.

rfk:
In service of https://bugzilla.mozilla.org/show_bug.cgi?id=1300451


rtilder:
r+

PR #42, opened 2016-10-04
add metrics.md

dannycoates:
This documents the metric data we collect from the sync server. r? @rfk @davismtl


rfk:
Whoops, I lost this in my inbox, taking a look now

metrics.md
```diff
@@ -0,0 +1,238 @@
...
+  code SMALLINT encode lzo,            -- http status code of the response
+  bucket VARCHAR(255) encode bytedict, -- sync bucket name (bookmarks, history, etc)
+  t INTEGER encode bytedict,           -- request time in milliseconds
+  ua_browser VARCHAR(255) encode lzo,  -- request User Agent browser
+  ua_version INTEGER encode lzo,       -- request User Agent browser version
+  ua_os VARCHAR(255) encode lzo,       -- request User Agent OS
+  host VARCHAR(255) encode lzo         -- server hostname that handled the request
+)
+distkey(uid)
+interleaved sortkey(uid, dev);
```
rfk: Out of curiosity, why sort by these instead of e.g. timestamp?

metrics.md
```diff
@@ -0,0 +1,238 @@
...
+    trunc(max(ts)) as day,
+    min(t) as min_t,
+    max(t) as max_t,
+    sum(posts) as posts,
+    sum(gets) as gets,
+    sum(puts) as puts,
+    sum(dels) as dels,
+    sum(aoks) as aoks,
+    sum(oops) as oops,
+    sum(fups) as fups,
```
rfk: I like these names :-P

Where are `fups`, `oops` etc defined?  I don't see them in the subquery.

metrics.md
```diff
@@ -0,0 +1,238 @@
...
+    case when bucket = 'tabs' and method = 'POST' then 1 end as w_tabs,
+    case when bucket = 'passwords' and method = 'POST' then 1 end as w_passwords,
+    case when bucket = 'addons' and method = 'POST' then 1 end as w_addons r_history,
+  from sync0715)
+group by uid, dev);
+```
+
+### Device Counts
+
+This table summarizes how many active devices each user had on a given day, where an active device is one that has had any activity in the last 7 days.
```
rfk: hey @davismtl, this one matches your request for "activity in last 7 days", so we should definitely go from 5 to 7 in FxA for consistency.

rfk:
Thanks @dannycoates, this LGTM overall, although it does feel like the sort of thing that could easily get out of date as the implementation changes.  Would it make sense to link to the live graphs in re:dash from here as well?

metrics.md
```diff
@@ -0,0 +1,238 @@
...
+
+## Artifacts
+
+Engagement and retention are our bread and butter metrics. By improving these measurements we can be confident that we're improving the user experience.
+
+### Engagement Ratio
+
+    Of all the people who used Sync in the last (28) days, what percentage of them used it on a given day?
+
+Having a high ratio of daily users in relation to monthly users  indicates that a majority of users are using Sync frequently. Tracking the engagement ratio allows us to monitor how changes to the user experience affect how frequently users engage.
```
davismtl: It might help to add an example. 

> For example, if we calculate an engagement ratio of 0.5, it means that users return to our product every 2 days. The calculation would look like 14 days / 28 days.
> Inversely, if users engage with our product on average 7 days per month, we will have an engagement ration of 0.25 (7/28).

metrics.md
```diff
@@ -0,0 +1,238 @@
...
+
+Here's the schema for the log tables:
+
+```sql
+create table sync0715
+(
+  uid CHAR(32) NOT NULL encode lzo,    -- a sha256-hashed Firefox Account (FxA) user id
+  s_uid CHAR(32) encode lzo,           -- a surrogate user id (generated by token server)
+  dev CHAR(32) NOT NULL encode lzo,    -- a sha256-hashed FxA device id
+  s_dev CHAR(32) encode lzo,           -- a surrogate device id (generated by token server)
```
davismtl: What are surrogate IDs? (both device and user)

metrics.md
davismtl: +1
I'm a bit confused here. :-/

metrics.md
davismtl: I think so. Good catch. :)

metrics.md
```diff
@@ -0,0 +1,238 @@
...
+)
+distkey(uid)
+interleaved sortkey(uid, dev);
+```
+
+## Rollups
+
+After a daily table has been loaded, a few sql commands do rollups of the data.
+
+### Device Activity
```
davismtl: Perhaps it's just me but I feel like it would help to see definitions of what each column in the table will look like. The queries (although very nice to have and more detailed than many teams) provide at times too much detail in order to quickly understand the available data.

Here are examples of how other teams have defined what data is available in there tables:
- Android: https://wiki.mozilla.org/Mobile/Metrics/Redash
- FxA funnel data: https://github.com/mozilla/fxa-auth-server/blob/7ef8d5d1c9e02f9bc119adf4b6356eb5716beb7d/docs/metrics-events.md

davismtl:
Great job Danny! Can't wait to jump into all this data. :)

PR #44, opened 2016-11-03
Batch upload ttl fixes

rfk:
This fixes the handling of TTL updates in a batch upload, allowing us to unskip the final two testcases for the batch upload work.  See https://bugzilla.mozilla.org/show_bug.cgi?id=1311903.

@Natim r?

It got a little thorny sorry, since it uncovered a bug in the way that the memcache layer was handling TTLs.  It may be easiest to review one commit at a time, and I'll leave some comments below.

The guts of the change is in the third commit, which updates the DB queries to provide proper partial-update semantics:

* If the batch upload did not specify a value of a field, keep any existing value
* Otherwise, use the new value from the batch upload

### DIFF
--- a/syncstorage/storage/sql/__init__.py
+++ b/syncstorage/storage/sql/__init__.py
@@ -523,6 +523,7 @@ class SQLStorage(SyncStorage):
             "userid": userid,
             "collection": collectionid,
             "default_ttl": MAX_TTL,
+            "ttl_base": int(session.timestamp),
             "modified": ts2bigint(session.timestamp)
         }
         session.query("APPLY_BATCH_UPDATE", params)
@@ -638,10 +639,6 @@ class SQLStorage(SyncStorage):
         row = {}
         row["userid"] = userid
         row["collection"] = collectionid
-        self._prepare_common_row(session, data, item, row)
-        return row
-
-    def _prepare_common_row(self, session, data, item, row):
         row["id"] = item
         if "sortindex" in data:
             row["sortindex"] = data["sortindex"]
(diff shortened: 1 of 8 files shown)


syncstorage/storage/sql/__init__.py
```diff
@@ -638,10 +639,6 @@ def _prepare_bso_row(self, session, userid, collectionid, item, data):
         row = {}
         row["userid"] = userid
         row["collection"] = collectionid
-        self._prepare_common_row(session, data, item, row)
-        return row
-
-    def _prepare_common_row(self, session, data, item, row):
```
rfk: It turns out that the row preparation is not that comment between the two queries after all :-)

syncstorage/storage/sql/dbconnect.py
```diff
@@ -114,28 +114,22 @@ def __new__(cls, *args, **kwds):
...
-def _common_columns(ttl):
-    return (
-        Column("sortindex", Integer),
-        Column("modified", BigInteger, nullable=False),
-        # I'd like to default these to the emptry string and zero,
-        # but MySQL doesn't let you set a default on a TEXT column...
-        Column("payload", PAYLOAD_TYPE, nullable=False),
-        Column("payload_size", Integer, nullable=False),
-        Column("ttl", Integer, server_default=sqltext(str(ttl)))
-    )
```
rfk: The "bso" and "batch_upload_items" tables actually need different column definitions - in the batch-upload case, we need to know whether any particular field was NULL.

syncstorage/views/__init__.py
```diff
@@ -461,7 +463,9 @@ def get_item(request):
     userid = request.validated["userid"]
     collection = request.validated["collection"]
     item = request.validated["item"]
-    return storage.get_item(userid, collection, item)
+    bso = storage.get_item(userid, collection, item)
+    bso.pop("ttl", None)
+    return bso
```
rfk: Although we return the "ttl" field so that it can be used by memcached, we don't want to return it to the client.  This change ensure it's removed at the network boundary.

syncstorage/tests/functional/test_storage.py
```diff
@@ -1485,7 +1484,6 @@ def test_batch_ttl_update(self):
         self.assertEquals(len(res), 1)
         self.assertEquals(res[0]['payload'], 'see')
 
-    @unittest2.skip("for the moment")
     def test_batch_ttl_is_based_on_commit_timestamp(self):
```
rfk: Aye, some sort of sporadic failure on mysql backend;  swear it was working on my machine :-)

rfk: I've adjusted the timings in this test to make it run more reliably (but slower).  It's passing well on my machine, let's try travis again...

Natim: Testing ttl's is always a pain. Would it be possible to use [freezegun](https://github.com/spulec/freezegun) instead of time.sleep?

Natim: Something like https://github.com/spulec/freezegun#manual-ticks

rfk: That's cool!  Part of the reason we use real times in these tests, is that we also run them against a live server to test for spec conformance.  But it may be possible to use the freezegun approach when running locally, and only sleep when running against a remote server.

PR #45, opened 2016-11-03
Add a feature-flag for the batch upload feature.

rfk:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1304627

This will let us roll out the new feature slowly, by only enabling it on nodes with the following config:

```
[storage]
batch_upload_enabled = true
```

It will be off by default.  Once we're happy that it's properly deployed in production, we can remove this and have it on all the time.

@Natim r?

### DIFF
--- a/syncstorage/tests/functional/test_storage.py
+++ b/syncstorage/tests/functional/test_storage.py
@@ -1677,6 +1677,89 @@ class TestStoragePaginated(TestStorage):
     TEST_INI_FILE = "tests-paginated.ini"
 
 
+class TestStorageWithBatchUploadDisabled(TestStorage):
+    """Storage testcases run with batch uploads disabled via feature flag."""
+
+    TEST_INI_FILE = "tests-no-batch.ini"
+
+    def test_batches(self):
+        # This is the same sequence of requests as the master
+        # test, but without the checks for batch semantics.
+        # It lets us know that batch stuff is properly ignored.
+
+        endpoint = self.root + '/storage/col2'
+
+        bso1 = {'id': '12', 'payload': 'elegance'}
+        bso2 = {'id': '13', 'payload': 'slovenly'}
+        bsos = [bso1, bso2]
+        self.app.post_json(endpoint, bsos)
+
+        bso3 = {'id': 'a', 'payload': 'internal'}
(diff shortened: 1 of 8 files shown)


rfk:
The intended semantics here is that when the feature is disabled, the batch-related query parameters are ignored.  Firefox code is currently shipping that expects it to work this way, e.g. it will pass `batch=true` as a query parameter, and when it sees that it didn't get given a batchid back, it will assume the server does not support this functionality.

We will of course have to actually test this in stage to make sure :-)

rfk:
Hmm, looks like tests not quite passing yet...

syncstorage/tests/tests-memcached-cacheonly.ini
```diff
@@ -11,6 +11,7 @@ backend = syncstorage.storage.memcached.MemcachedStorage
 wraps = sqlstorage
 cache_key_prefix = sync-${MOZSVC_UUID}-
 cache_only_collections = meta tabs col1 col2
+batch_upload_enabled = true
```
Natim: Does it means that it is disabled by default?

Natim: Yes it default to False below.

syncstorage/tests/functional/test_storage.py
```diff
@@ -1677,6 +1677,83 @@ class TestStoragePaginated(TestStorage):
...
+        resp = self.app.get(endpoint + '/13')
+        self.assertEquals(resp.json['payload'], 'portnoy')
+        self.assertEquals(committed, float(resp.headers['X-Last-Modified']))
+        self.assertEquals(committed, resp.json['modified'])
+        resp = self.app.get(endpoint + '/c')
+        self.assertEquals(resp.json['payload'], 'tinsel')
+        self.assertEquals(committed, resp.json['modified'])
+        resp = self.app.get(endpoint + '/14')
+        self.assertEquals(resp.json['payload'], 'itsybitsy')
+        self.assertEquals(committed, resp.json['modified'])
```
Natim: Should we validate there that we don't receive any batch id?

rfk:
Thanks @Natim!  I've made some tweaks based on your feedback.

PR #46, opened 2016-11-03
Enable and change behaviour of disabled batch tests

mostlygeek:
- enable test_batch_ttl_update
- enable test_batch_ttl_is_based_on_commit_timestamp
- change to check for 200 status instead of 202 (Accepted) from
  batch API

Let's discuss the change from 202 (Accepted) to 200. I think 200 are actually the right response code for the API to return. A 202 responses means the request was accepted but has not finished processing. That's not accurate because the data has already been written to disk and there nothing further to do. Committing the batch will be another discrete operation. 

@rtilder @rfk r?

### DIFF
--- a/syncstorage/tests/functional/test_storage.py
+++ b/syncstorage/tests/functional/test_storage.py
@@ -1448,7 +1448,6 @@ class TestStorage(StorageFunctionalTestCase):
         # Fields not touched by the batch, should have been preserved.
         self.assertEquals(res[1]['sortindex'], 17)
 
-    @unittest2.skip("for the moment")
     def test_batch_ttl_update(self):
         collection = self.root + '/storage/col2'
         bsos = [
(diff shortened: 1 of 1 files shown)


rfk:
This is going to conflict with https://github.com/mozilla-services/server-syncstorage/pull/44

rfk:
> Let's discuss the change from 202 (Accepted) to 200.
> I think 200 are actually the right response code for the API to return

/cc @mhammond, but I think that the Firefox client code already does some sort of explicit check for the 202 response code, so it may be complicated to change.

kitcambridge:
It looks like Firefox does explicitly check for 202: http://searchfox.org/mozilla-central/rev/f5c9e9a249637c9abd88754c8963ecb3838475cb/services/sync/modules/record.js#942-949

mostlygeek:
Ahh! if FF checks for 202 then I think it'll be better to just close this PR :)

PR #48, opened 2016-11-03
Add tests to ensure users can't write to each other's batches.

rfk:
The tests pass without any changes to the code, but it's nice
to have them explicit, especially since the batchid provides
a layer of indirection when writing to a collection.

@Natim r?


rfk:
> The tests pass without any changes to the code

Heh, actually, looks like it surfaced a bug in the memcache layer!

PR #49, opened 2016-11-03
Make batch and non-batch posts have same semantics for invalid items.

rfk:
In the non-batch case, we ignore individual invalid items and continue with processing the valid ones.  Since we want the semantics of batch uploads to be as close to non-batch as possible, we should apply the same rules there.

For a fun bonus, this uncovered an edge-case where we'd incorrectly accept a BSO id with a trailing newline, because regexes.


rfk:
@Natim r?

PR #51, opened 2016-11-04
Fix test_batch_with_failing_bsos commit logic

mostlygeek:
When the batch is committed the success criteria should be be 2
successful BSOs not 1. BSOs ["a", "c"] are committed so the "success" key
should account for both of them.

@rfk r? Please

### DIFF
--- a/syncstorage/tests/functional/test_storage.py
+++ b/syncstorage/tests/functional/test_storage.py
@@ -1575,7 +1575,7 @@ class TestStorage(StorageFunctionalTestCase):
         endpoint = collection + '?batch={0}&commit=true'.format(batch)
         resp = self.app.post_json(endpoint, bsos)
         self.assertEqual(len(resp.json['failed']), 1)
-        self.assertEqual(len(resp.json['success']), 1)
+        self.assertEqual(len(resp.json['success']), 2)
 
         # To correctly match semantics of batchless POST, the batch
         # should be committed including only the successful items.


syncstorage/tests/functional/test_storage.py
```diff
@@ -1575,7 +1575,7 @@ def test_batch_with_failing_bsos(self):
         endpoint = collection + '?batch={0}&commit=true'.format(batch)
         resp = self.app.post_json(endpoint, bsos)
         self.assertEqual(len(resp.json['failed']), 1)
-        self.assertEqual(len(resp.json['success']), 1)
+        self.assertEqual(len(resp.json['success']), 2)
```
rfk: The "a" item was already reported as successful in the previous POST.  If we include it here as well, the semantics of the "success" list change between non-batch uploads (where it's only items from the request body) and batch uploads (where it would be any items from the batch).

mostlygeek: I think what "success" means needs some clarity. I would expect to interpret the results like this: 

First Request: 

- `a` was successfully appended to the batch
- `b` failed and reasons are listed in `failed`. In this case, with a bad id

The response tells me what was successful/failed to be appended to a batch.

Second Request:

- `a` was successfully committed
- `c` was successfully committed
- `d` failed due to reasons and was not committed 

With a `commit=true` the response tells me what was successfully committed to the database. The data is now officially in the collections.

The ambiguity would make it harder to implement clients correctly. If Request #1 sends a success and then the batch times out and is purged for some reason what does that mean to the client? 

If there's already client code that is "done", I'm willing to acquiesce to that behavior if I have to. :)

rfk: ISTM that it should not be possible for `a` to be accepted into the batch, but then later fail to commit as part of the batch.  Either the *entire* commit failed and you get a non-200 response, or it succeeded and included all the things that were successfully added to it.  So I don't really seem any ambiguity, but I could easily be missing something.

> If Request #1 sends a success and then the batch times out and
> is purged for some reason what does that mean to the client? 

It can't assume that `a` will be committed until it has committed the batch, but it's not obvious how including `a` in the final "success" list makes that any easier to keep track of.

I'm tentatively open to changing it if this makes more sense to the client, but would like to get Mark or Thom's input before committing either way.

mostlygeek:
Closing this PR. We can always revisit in the future if requirements from the client side dictate.

PR #52, opened 2016-11-04
Address miscellaneous review nits for batch-uploads work.

rfk:
This is, I hope, my last suite of changes against the batch-uploads branch.  It addresses a bunch of smaller nits I had in my initial review, about which I'll leave comments throughout.  @Natim r?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,14 +8,12 @@ BUILD_TMP = /tmp/syncstorage-build.${USER}
 PYPI = https://pypi.python.org/simple
 
 export MOZSVC_SQLURI = sqlite:///:memory:
-#export MOZSVC_SQLURI = mysql://user:password@localhost/my_sync_db
 
 # Hackety-hack around OSX system python bustage.
 # The need for this should go away with a future osx/xcode update.
 ARCHFLAGS = -Wno-error=unused-command-line-argument-hard-error-in-future
 
 INSTALL = ARCHFLAGS=$(ARCHFLAGS) $(PIP) install -U -i $(PYPI)
-
 
 .PHONY: all build test
 
--- a/loadtest/stress.py
+++ b/loadtest/stress.py
@@ -178,14 +178,8 @@ class StressTest(TestCase):
 
         # POST requests with several WBOs batched together
         num_requests = self._pick_weighted_count(post_count_distribution)
-        # Shallower distribution(for now) of "transactional" style batch
-        # uploads.  Looking for roughly 10% of batch POSTs to be
-        # "transactional".  Increase the frequency of "transactional" requests
-        # by changing the modulo 10 to something like modulo 5(~20%) or
-        # modulo 3(~30%).
-        # transact = int(random.paretovariate(1) * 100) % 10 == 0
-        #
-        # Let's do roughly 50% for fun.
+
+        # Let's do roughly 50% transactional batches.
         transact = random.randint(0, 1)
         batch_id = None
         committing = False
--- a/syncstorage/storage/__init__.py
+++ b/syncstorage/storage/__init__.py
@@ -16,6 +16,9 @@ from mozsvc.plugin import resolve_name
 
 logger = logging.getLogger("syncstorage.storage")
 
+# Rough guesstimate of the maximum reasonable life span of a batch.
+BATCH_LIFETIME = 60 * 60 * 2  # 2 hours, in seconds
+
 
 class StorageError(Exception):
     """Base class for exceptions raised from the storage backend."""
--- a/syncstorage/storage/memcached.py
+++ b/syncstorage/storage/memcached.py
@@ -59,7 +59,8 @@ from syncstorage.storage import (SyncStorage,
                                  CollectionNotFoundError,
                                  ItemNotFoundError,
                                  InvalidOffsetError,
-                                 InvalidBatch)
+                                 InvalidBatch,
+                                 BATCH_LIFETIME)
 
 from pyramid.settings import aslist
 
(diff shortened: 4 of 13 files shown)


Makefile
```diff
@@ -8,15 +8,13 @@ BUILD_TMP = /tmp/syncstorage-build.${USER}
...
 export MOZSVC_SQLURI = sqlite:///:memory:
-#export MOZSVC_SQLURI = mysql://user:password@localhost/my_sync_db
 
 # Hackety-hack around OSX system python bustage.
 # The need for this should go away with a future osx/xcode update.
 ARCHFLAGS = -Wno-error=unused-command-line-argument-hard-error-in-future
 
 INSTALL = ARCHFLAGS=$(ARCHFLAGS) $(PIP) install -U -i $(PYPI)
 
-
```
rfk: These changes were noise in the original PR, restoring them to their original values.

loadtest/stress.py
```diff
@@ -178,17 +178,12 @@ def test_storage_session(self):
...
         # POST requests with several WBOs batched together
         num_requests = self._pick_weighted_count(post_count_distribution)
-        # Shallower distribution(for now) of "transactional" style batch
-        # uploads.  Looking for roughly 10% of batch POSTs to be
-        # "transactional".  Increase the frequency of "transactional" requests
-        # by changing the modulo 10 to something like modulo 5(~20%) or
-        # modulo 3(~30%).
-        # transact = int(random.paretovariate(1) * 100) % 10 == 0
-        #
-        # Let's do roughly 50% for fun.
```
rfk: Remove commented-out code.

syncstorage/storage/__init__.py
```diff
@@ -16,6 +16,9 @@
 
 logger = logging.getLogger("syncstorage.storage")
 
+# Rough guesstimate of the maximum reasonable life span of a batch.
+BATCH_LIFETIME = 60 * 60 * 2  # 2 hours, in seconds
+
```
rfk: I made this a sharable constant, previously it was hard-coded in several places.

syncstorage/storage/memcached.py
```diff
@@ -944,23 +945,31 @@ def del_item(self, userid, item):
...
+            ts = get_timestamp()
+        ts = int(ts)
+        bdata, bcasid = self.cache.gets(self.get_batches_key(userid))
+        # Remove any expired batches, but let the
+        # calling code write it back out to memcache.
+        if bdata:
+            for batchid, batch in bdata.items():
+                if batch["created"] + BATCH_LIFETIME < ts:
+                    del bdata[batchid]
+        return bdata, bcasid
```
rfk: This consolidates the logic for expiring batches from memcache in a single location.

syncstorage/storage/memcached.py
```diff
@@ -944,23 +945,31 @@ def del_item(self, userid, item):
...
-        bdata[batchid] = {"user": userid,
-                          "modified": batch,
-                          # FIXME Rough guesstimate of the maximum
-                          #       reasonable life span of a batch
-                          "expires": int(batch) + 2 * 3600,
-                          "items": []}
+        bdata[batchid] = {
+            "created": int(ts),
+            "items": []
+        }
```
rfk: The "user" and "modified" keys were never used, so I removed them to save space.  Also changed "expires" to "created" in case we want to make the batch lifetime configurable in future.

syncstorage/storage/sql/__init__.py
```diff
@@ -502,6 +507,7 @@ def valid_batch(self, session, userid, collection, batchid):
         valid = session.query_scalar("VALID_BATCH", params=params)
         return valid
 
+    @metrics_timer("syncstorage.storage.sql.append_items_to_batch")
```
rfk: Add another timer for monitoring in prod.

syncstorage/storage/sql/queries_generic.py
```diff
@@ -251,14 +251,19 @@ def FIND_ITEMS(bso, params):
...
 # Unfortunately there's no generic way to achieve this in SQL so the default
 # case winds up deleting all expired items.  There is a MySQL-specific
-# version using DELETE <blah> LIMIT 1000.
-PURGE_SOME_EXPIRED_ITEMS = "DELETE FROM %(bso)s "\
-                           "WHERE ttl < (UNIX_TIMESTAMP() - :grace) " \
-                           "LIMIT :maxitems"
+# version using DELETE <ttlblah> LIMIT 1000.
 
-PURGE_BATCHES = "DELETE FROM batch_uploads WHERE batch < " \
-                "   (UNIX_TIMESTAMP() - :grace) * 1000 LIMIT :maxitems"
```
rfk: These queries were not in fact generic, I've changed them to not use the mysql-specific "LIMIT" clause.

syncstorage/storage/sql/queries_mysql.py
```diff
@@ -10,13 +10,23 @@
...
-                           "ORDER BY ttl LIMIT :maxitems"
+PURGE_SOME_EXPIRED_ITEMS = """
+    DELETE FROM %(bso)s
+    WHERE ttl < (UNIX_TIMESTAMP() - :grace)
+    ORDER BY ttl LIMIT :maxitems
+"""
+
+PURGE_BATCHES = """
+    DELETE FROM batch_uploads
+    WHERE batch < (UNIX_TIMESTAMP() - :lifetime - :grace) * 1000
```
rfk: Note that we now pass in the batch lifetime as a separate parameter, rather than hard-coding it into the calculation of the grace period.  This has further simplified the tests for this code.

syncstorage/storage/sql/queries_sqlite.py
```diff
@@ -22,14 +22,20 @@
...
+PURGE_SOME_EXPIRED_ITEMS = """
+    DELETE FROM %(bso)s
+    WHERE ttl < (strftime('%%s', 'now') - :grace)
+"""
 
-PURGE_BATCHES = "DELETE FROM batch_uploads WHERE batch < " \
-                "   (SELECT strftime('%%s', 'now') - :grace) * 1000"
+PURGE_BATCHES = """
+    DELETE FROM batch_uploads
+    WHERE batch < (strftime('%s', 'now') - :lifetime - :grace) * 1000
```
rfk: There was a subtle bug here, the '%%s' was incorrect because this string is never run through the string formatter, unlike the other two which include '%(bui)s' table references.  It should not escape the percent sign.

syncstorage/storage/sql/__init__.py
```diff
@@ -514,7 +520,7 @@ def append_items_to_batch(self, session, userid, collection, batchid,
         session.insert_or_update("batch_upload_items", rows)
         return session.timestamp
 
-    @metrics_timer("syncstorage.storage.sql.db.execute.apply_batch")
+    @metrics_timer("syncstorage.storage.sql.apply_batch")
```
Natim: Do we test that the `metrics_timer` applies on the proper function?

rfk: I don't think so, but you're right that we should.

rfk:
Thanks @Natim, I've added a test that we're emitting the right metrics timers for batch uploads, based on a similar timer that was already in place.  Will merge once CI goes green.

PR #54, opened 2016-11-23
Skip user-switching functional tests when running them live.

rfk:
This issue was causing test failures when updating https://github.com/mozilla-services/syncserver to the latest tag.  Since it's only tests and lint, merging r=myself to fix the tests.


rfk:
> merging r=myself to fix the tests

(but not until travis passes here, obviously)

PR #55, opened 2017-01-11
Bug 1324600 - test that batch writes correctly report X-Last-Modified

rfk:


rfk:
@mostlygeek reports these working as intended over in https://bugzilla.mozilla.org/show_bug.cgi?id=1324600#c17, merging

PR #56, opened 2017-01-20
Add a functional test for expected behaviour when ttl=null

mostlygeek:
Ref: [Bug 1332552](https://bugzilla.mozilla.org/show_bug.cgi?id=1332552). iOS clients send a ttl:null for BSOs in the JSON. This
is meant to be interpreted as never expire. The gosync server doesn't
behave correctly with ttl=null. Add a test to prevent regressions.

@rfk r? please

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -16,18 +16,17 @@ dependencies:
     # Create a version.json file for the app to serve; copy it to the CircleCI
     # artifacts directory for debugging.
     - >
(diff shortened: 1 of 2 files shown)


mostlygeek:
Failed some styling. `make test` successfully locally. 

@rfk could you merge this and tag a 1.6.2 release. Then I'll update gosync's CI to use 1.6.2 which will test for ttl:null.

rfk:
Thanks, I've tagged a `1.6.2` release.

mostlygeek:
@rfk I pushed the tag. Thanks!

PR #59, opened 2017-03-08
Reject in-flight batches when batch uploads are disabled.

rfk:
If we silently accept them, we'll do the wrong thing from what the client expects.


syncstorage/tests/functional/test_storage.py
```diff
@@ -1921,6 +1921,11 @@ def test_batches(self):
         bso6 = {'id': '13', 'payload': 'portnoy'}
         bso0 = {'id': '14', 'payload': 'itsybitsy'}
         commit = '?batch={0}&commit=true'.format(batch)
+        # This errors out because it's trying to use an in-flight batch.
+        resp = self.app.post_json(endpoint + commit, [bso5, bso6, bso0],
+                                  status=400)
+        # This reqeusts a new batch, which is silently ignored and succeeds.
```
rnewman: requests

PR #60, opened 2017-03-08
Bug 1342679 - Properly track timestamps of updated items in memcached.

rfk:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1342679 for reference.

There's a bug in the memcache storage layer where, if you update some existing BSOs but don't add any new ones, then the cached collection data is not updated with the new "modified" timestamp.  Anyone that looks at this cached data to determine the modified time of the collection, will think it's unmodified when it actually is.

AFAICT this bug has been here for years, but we haven't noticed it because there's an additional layer of caching that contains the timestamps for all the collections.  That layer *was* getting updated correctly, so there was never much opportunity to observe the incorrect timestamp:

1. key `<userid>:metadata` contains the last-modified time for all collections, and was being updated correctly
2. key `<userid>:c:<collection`  contains cached data for a specific collection, and its modified time was not getting updated correctly

However!  With the new batch upload API, we started leaving data out of (1) and instead falling back to the data in (2).  This causes us to start observing incorrect last-modified times for the "tabs" collection, which is stored solely in memcache.

It *may* also have been possible to observe incorrect last-modified times for the "clients" and "meta" collections, which are stored both in memcache and the db.  I wasn't able to reproduce this in tests however.

### DIFF
--- a/syncstorage/storage/memcached.py
+++ b/syncstorage/storage/memcached.py
@@ -369,17 +369,22 @@ class MemcachedStorage(SyncStorage):
     def append_items_to_batch(self, userid, collection, batchid, items):
         """Appends items to the pending batch."""
         colmgr = self._get_collection_manager(collection)
-        ts = colmgr.append_items_to_batch(userid, batchid, items)
-        return ts
+        with self._mark_collection_dirty(userid, collection) as update:
+            ts = colmgr.append_items_to_batch(userid, batchid, items)
+            # Account for the size of the new items as they come in,
+            # since that's the only opportunity we have to see them.
+            # Don't update the timestamp yet though, as they're not committed.
+            size = sum(len(item.get("payload", "")) for item in items)
+            update(size_incr=size)
+            return ts
 
     def apply_batch(self, userid, collection, batchid):
         """Applies the batch"""
         colmgr = self._get_collection_manager(collection)
-        # prevts = colmgr.get_timestamp(userid)
-        with self._mark_collection_dirty(userid, collection):
+        with self._mark_collection_dirty(userid, collection) as update:
             ts = colmgr.apply_batch(userid, batchid)
-            # Lazy update should occur on following get_item/s
-        return ts
+            update(ts, ts)
+            return ts
 
     def close_batch(self, userid, collection, batchid):
         colmgr = self._get_collection_manager(collection)
@@ -762,7 +767,7 @@ class _CachedManagerBase(object):
                 if "payload" not in bso:
                     bso["payload"] = ""
                 data["items"][bso["id"]] = bso
-                data["modified"] = modified
+            data["modified"] = modified
         # Purge any items that have expired.
         # We can't do this as part of the purge_expired_items()
         # because we don't have a way to enumerate all user ids.
--- a/syncstorage/tests/functional/test_storage.py
+++ b/syncstorage/tests/functional/test_storage.py
@@ -72,17 +72,32 @@ class TestStorage(StorageFunctionalTestCase):
         finally:
             self.root = orig_root
 
-    def test_get_collections(self):
+    def test_get_info_collections(self):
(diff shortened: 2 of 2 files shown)


syncstorage/storage/memcached.py
```diff
@@ -369,17 +369,22 @@ def valid_batch(self, userid, collection, batchid):
...
         """Applies the batch"""
         colmgr = self._get_collection_manager(collection)
-        # prevts = colmgr.get_timestamp(userid)
-        with self._mark_collection_dirty(userid, collection):
+        with self._mark_collection_dirty(userid, collection) as update:
             ts = colmgr.apply_batch(userid, batchid)
-            # Lazy update should occur on following get_item/s
-        return ts
+            update(ts, ts)
+            return ts
```
rfk: Here, we were chosing not to update the timestamp in the `<userid>:metadata` collection, and instead to lazily fall back to reading it from the cached collection data after applying a batch.  This is the reason why we started observing this bug when the batch upload API was enabled.

syncstorage/storage/memcached.py
```diff
@@ -762,7 +767,7 @@ def _set_items(self, userid, items, modified, data, casid):
                 if "payload" not in bso:
                     bso["payload"] = ""
                 data["items"][bso["id"]] = bso
-                data["modified"] = modified
+            data["modified"] = modified
```
rfk: This is the buggy line of code - we should update the "modified" timestamp whenever any items change, but this indentation meant it was only updated when you add new items.

syncstorage/tests/functional/test_storage.py
```diff
@@ -72,17 +72,32 @@ def _switch_user(self):
...
+        bsos = [{"id": str(i), "payload": "yyy"} for i in xrange(2)]
+        resp = self.app.post_json(self.root + "/storage/col2", bsos)
+        self.assertTrue(ts2 < resp.json["modified"])
+        ts2 = resp.json["modified"]
+        resp = self.app.get(self.root + '/info/collections')
+        res = resp.json
+        keys = sorted(res.keys())
+        self.assertEquals(keys, ["col1", "col2"])
+        self.assertEquals(res["col1"], ts1)
+        self.assertEquals(res["col2"], ts2)
```
rfk: This was some extra sanity-checking I added to ensure /info/collections behaves as intended.

syncstorage/tests/functional/test_storage.py
```diff
@@ -1636,21 +1656,69 @@ def test_batch_with_immediate_commit(self):
           {'id': 'b', 'payload': 'bie'},
           {'id': 'c', 'payload': 'cee'}
         ]
-        resp = self.app.post_json(collection, bsos)
```
rfk: This was a totally pointless additional post, which happened to trigger the bug!  I've moved it into a separate testcase for clarity.

syncstorage/tests/functional/test_storage.py
```diff
@@ -1636,21 +1656,69 @@ def test_batch_with_immediate_commit(self):
...
         resp = self.app.get(collection + '?full=1')
+        self.assertEquals(float(resp.headers['X-Last-Modified']), committed)
         res = resp.json
         res.sort(key=lambda bso: bso['id'])
         self.assertEquals(len(res), 3)
         self.assertEquals(res[0]['payload'], 'aih')
         self.assertEquals(res[1]['payload'], 'bie')
         self.assertEquals(res[2]['payload'], 'cee')
 
+    def test_batch_uploads_properly_update_info_collections(self):
```
rfk: And here's the new test case that exhibited this bug.

syncstorage/storage/memcached.py
mostlygeek: brutal. :)

rfk:
Thanks for speedy review @Natim @mostlygeek!

PR #62, opened 2017-06-01
Add functional test for delete collection then post scenario

mostlygeek:
A bug was discovered in gosync (issue #169) where a DELETE of a
collection followed by a POST had the wrong behaviour. Add
test_delete_collection_then_post to ensure all servers conform to the
expected behaviour by clients.

### DIFF
--- a/syncstorage/tests/functional/test_storage.py
+++ b/syncstorage/tests/functional/test_storage.py
@@ -574,6 +574,64 @@ class TestStorage(StorageFunctionalTestCase):
         res = self.app.get(self.root + '/storage/col2')
         self.assertEquals(len(res.json), 0)
 
+    def test_delete_collection_then_post(self):
+        # creating a collection of three
+        bso1 = {'id': '12', 'payload': _PLD}
+        bso2 = {'id': '13', 'payload': _PLD}
+        bso3 = {'id': '14', 'payload': _PLD}
+        bsos = [bso1, bso2, bso3]
+
+        self.app.post_json(self.root + '/storage/col2', bsos)
+        res = self.app.get(self.root + '/storage/col2')
+        self.assertEquals(len(res.json), 3)
+
+        resp = self.app.get(self.root + '/info/collections')
+        res = resp.json
+        self.assertTrue("col2" in res)
+
+        infoBefore = self.app.get(self.root + '/info/collections')
+
+        # Deleting the collection should result in:
+        #  - collection does not appear in /info/collections
+        #  - X-Last-Modified timestamp at the storage level changing
+        self.app.delete(self.root + '/storage/col2')
+        items = self.app.get(self.root + '/storage/col2').json
+        self.assertEquals(len(items), 0)
+
+        infoAfter1 = self.app.get(self.root + '/info/collections')
+        res = infoAfter1.json
+        self.assertTrue("col2" not in res)
+        self.assertTrue("X-Last-Modified" in resp.headers)
+        self.assertNotEqual(infoBefore.headers["X-Last-Modified"],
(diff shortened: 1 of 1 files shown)


syncstorage/tests/functional/test_storage.py
```diff
@@ -574,6 +574,63 @@ def test_delete_collection_items(self):
...
+
+        self.app.post_json(self.root + '/storage/col2', bsos)
+        res = self.app.get(self.root + '/storage/col2')
+        self.assertEquals(len(res.json), 3)
+
+        resp = self.app.get(self.root + '/info/collections')
+        res = resp.json
+        self.assertTrue("col2" in res)
+
+        infoBefore = self.app.get(self.root + '/info/collections')
```
Natim: How is this different from the previous resp on L588?

syncstorage/tests/functional/test_storage.py
```diff
@@ -574,6 +574,63 @@ def test_delete_collection_items(self):
...
+
+        # make sure the storage level timestamp stays the same
+        time.sleep(0.2)
+        infoAfter2 = self.app.get(self.root + '/info/collections')
+        self.assertEqual(infoAfter1.headers["X-Last-Modified"],
+                         infoAfter2.headers["X-Last-Modified"])
+
+        # post BSOs again with X-I-U-S == 0
+        self.app.post_json(
+                self.root + '/storage/col2', bsos,
```
Natim: nit: This could fit on the previous line

syncstorage/tests/functional/test_storage.py
```diff
@@ -574,6 +574,63 @@ def test_delete_collection_items(self):
...
+        infoAfter3 = self.app.get(self.root + '/info/collections')
+        self.assertNotEqual(infoAfter3.headers["X-Last-Modified"],
+                            infoAfter2.headers["X-Last-Modified"])
+
+        # delete and POST again with X-I-U-S > 0
+        self.app.delete(self.root + '/storage/col2')
+        items = self.app.get(self.root + '/storage/col2').json
+        self.assertEquals(len(items), 0)
+        self.app.post_json(
+                self.root + '/storage/col2', bsos,
```
Natim: nit: This could fit on the previous line

syncstorage/tests/functional/test_storage.py
```diff
@@ -574,6 +574,63 @@ def test_delete_collection_items(self):
...
+
+        infoBefore = self.app.get(self.root + '/info/collections')
+
+        # Deleting the collection should result in:
+        #  - collection does not appear in /info/collections
+        #  - X-Last-Modified timestamp at the storage level changing
+        self.app.delete(self.root + '/storage/col2')
+        items = self.app.get(self.root + '/storage/col2').json
+        self.assertEquals(len(items), 0)
+
```
Natim: We should probably add a ``time.sleep`` here too in order to fix the test failures.

rfk:
> This LGTM apart from test nits, I'll push some tweaks and merge.

Oh snap, I think adding a strict ordering check in the tests has uncovered a bug in the python server.  I will see about a fix and push for re-review.

rfk:
The python server is doing this to determine the overall storage timestamp:

```
STORAGE_TIMESTAMP = "SELECT MAX(last_modified) FROM user_collections "\
                    "WHERE userid=:userid"
```
Which is obviously wrong, for the same reason that just taking the max modified time of a BSO is wrong for a collection.

rfk:
@bbangert I'd be curious whether the new spanner backend can or could pass this additional functional test, which the python backend currently doesn't.

bbangert:
Flagging this for @pjenvey lest we forget it forever.

PR #63, opened 2017-06-15
Prevent data corruption caused by batch id re-use.

rfk:
WIP, for now this is just a failing test...

### DIFF
--- a/syncstorage/storage/sql/__init__.py
+++ b/syncstorage/storage/sql/__init__.py
@@ -493,7 +493,11 @@ class SQLStorage(SyncStorage):
             "userid": userid,
             "collection": collectionid
         }
-        session.query("CREATE_BATCH", params)
+        try:
+            session.query("CREATE_BATCH", params)
+        except IntegrityError:
+            # The user tried to create two batches with the same timestamp.
+            raise ConflictError
         return batchid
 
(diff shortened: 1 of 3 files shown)


mostlygeek:
gosync passes the new functional test.

rfk:
@mostlygeek r?

syncstorage/storage/sql/__init__.py
```diff
@@ -493,7 +493,11 @@ def create_batch(self, session, userid, collection):
             "userid": userid,
             "collection": collectionid
         }
-        session.query("CREATE_BATCH", params)
+        try:
+            session.query("CREATE_BATCH", params)
+        except IntegrityError:
+            # The user tried to create two batches with the same timestamp.
```
mostlygeek: will this catch duplicate timestamps from different users ?

rfk: no; the primary key on this table is (batchid, userid) so it'll only trigger for duplicates of this tuple.

(This change isn't critical to the issue at hand, but was motivated by the new testcase.  It was possible for a user to trigger a `500 Server Error` by creating a batch and then quickly creating another without closing the first.  Now that will produce a `409 Conflict` which is more appropriate)

PR #65, opened 2017-06-16
Bug 1364005 - Reserve new standard collection ids for formautofill data.

rfk:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1364005


rfk:
It allegedly makes the standard collections a little more efficient, but TBH I don't think we ever measured the impact of it all.  Mostly it makes it easier for bob to look up standard collections in the db if we know what number they map to :-)

PR #66, opened 2017-07-11
Add pyup configuration.

Micheletto:
@rfk r?

### DIFF
--- a/.pyup.yml
+++ b/.pyup.yml
@@ -1,0 +1,6 @@
+schedule: "every day"
+search: False
+update: insecure
+requirements:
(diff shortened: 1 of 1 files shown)


.pyup.yml
```diff
@@ -0,0 +1,6 @@
+schedule: "every day"
+search: False
+update: insecure
```
rfk: o_O

@Micheletto what are the implications of "insecure" here?

Micheletto: That's for security updates only: https://pyup.io/docs/configuration/#control-whats-being-updated

PR #70, opened 2017-09-05
Bug 300451 - increase max payload size to 2MB

rfk:


syncstorage/storage/sql/dbconnect.py
```diff
@@ -108,8 +108,7 @@ def __new__(cls, *args, **kwds):
 # It is used to create either sharded or non-shareded BSO storage tables,
 # depending on the run-time settings of the application.
 
-PAYLOAD_TYPE = Text(length=256*1024)
```
rfk: Setting a max length on a text field is not supported by all db engines anyway, so I figure it's simpler to remove it here and reduce duplication of this magic value.

mostlygeek: We don't control all of the clients that talk to sync so we should enforce a limit.

PR #72, opened 2017-09-06
Empty request body should be valid applicatiln/newlines input

rfk:
This includes the tests from #71 and a code fix to make them pass.


rfk:
Oh hrm, got a test failure locally that I wasn't expecting, sec...

rfk:
> Oh hrm, got a test failure locally that I wasn't expecting, sec...

Aha, have to skip this test when batch uploads are disabled

PR #74, opened 2017-09-21
Ensure all size limits are inclusive, and are internally consistent.

### LINKED ISSUE #73: Inconsistency between BATCH_MAX_BYTES and BATCH_MAX_COUNT
As @thomcc found in https://bugzilla.mozilla.org/show_bug.cgi?id=1393659, BATCH_MAX_COUNT is, as its name implies, an "inclusive" max - the count can be equal to the max. However, BATCH_MAX_BYTES is exclusive - if the number of bytes is equal to that value, the BSO will be rejected.

These should be consistent. IMO, BATCH_MAX_BYTES should be made inclusive.

https://github.com/mozilla-services/server-syncstorage/blob/master/syncstorage/views/validators.py#L306-L313

(Note that in that bug we are accounting for this behaviour, so this bug isn't blocking desktop. However, it's not clear other implementations take this into account, nor that future implementations will)

rfk:
Fixes #73, adds explicit functional tests to assert https://bugzilla.mozilla.org/show_bug.cgi?id=1401707, and includes "max_request_bytes" in the configuration report even when batches are enabled.

@mhammond r?  The changes here are mostly to test cases so if you could check whether they make sense from your perspective that'd be awesome.

### DIFF
--- a/syncstorage/tests/functional/test_old_storage.py
+++ b/syncstorage/tests/functional/test_old_storage.py
@@ -20,7 +20,6 @@ from decimal import Decimal
 
 from syncstorage.tests.functional.support import StorageFunctionalTestCase
 from syncstorage.tests.functional.support import run_live_functional_tests
-from syncstorage.views.util import get_limit_config
 
 WEAVE_INVALID_WBO = 8
 
@@ -411,32 +410,6 @@ class TestOldStorage(StorageFunctionalTestCase):
         time.sleep(2.1)
         res = self.app.get(self.root + '/storage/col2')
         self.assertEquals(len(res.json), 0)
-
-    def test_batch(self):
-        max_count = get_limit_config(self.config, "max_post_records")
-        max_bytes = get_limit_config(self.config, "max_post_bytes")
-        self.assertEquals(max_bytes, 1024 * 1024)
-        # Test that batch uploads are correctly processed.
-        # Uploading max_count-5 small objects should succeed.
-        wbos = [{'id': str(i), 'payload': 'X'} for i in range(max_count - 5)]
-        res = self.app.post_json(self.root + '/storage/col2', wbos)
-        res = res.json
-        self.assertEquals(len(res['success']), max_count - 5)
-        self.assertEquals(len(res['failed']), 0)
-        # Uploading max_count+5 items should produce five failures.
-        wbos = [{'id': str(i), 'payload': 'X'} for i in range(max_count + 5)]
-        res = self.app.post_json(self.root + '/storage/col2', wbos)
-        res = res.json
-        self.assertEquals(len(res['success']), max_count)
-        self.assertEquals(len(res['failed']), 5)
-        # The test config has max_bytes=1M.
-        # Uploading 5 210MB items should produce one failure.
-        wbos = [{'id': str(i), 'payload': "X" * (210 * 1024)}
-                for i in range(5)]
-        res = self.app.post_json(self.root + '/storage/col2', wbos)
-        res = res.json
-        self.assertEquals(len(res['success']), 4)
-        self.assertEquals(len(res['failed']), 1)
 
     def test_weird_args(self):
         # pushing some data in col2
--- a/syncstorage/tests/functional/test_storage.py
+++ b/syncstorage/tests/functional/test_storage.py
@@ -779,16 +779,20 @@ class TestStorage(StorageFunctionalTestCase):
         res = self.app.get(self.root + '/storage/col2')
         self.assertEquals(len(res.json), 0)
 
-    def test_batch(self):
-        # This can't be run against a live server
-        # due to request size limits in nginx and
-        # the fact that it reads config variables.
-        if self.distant:
-            raise unittest2.SkipTest
+    def test_multi_item_post_limits(self):
+        res = self.app.get(self.root + '/info/configuration')
+        try:
+            max_bytes = res.json['max_post_bytes']
+            max_count = res.json['max_post_records']
+        except KeyError:
+            # Can't run against live server if it doesn't
+            # report the right config options.
+            if self.distant:
+                raise unittest2.SkipTest
+            max_bytes = get_limit_config(self.config, 'max_post_bytes')
+            max_count = get_limit_config(self.config, 'max_post_records')
 
-        # Test that batch uploads are correctly processed.
         # Uploading max_count-5 small objects should succeed.
-        max_count = get_limit_config(self.config, 'max_post_records')
         bsos = [{'id': str(i), 'payload': 'X'} for i in range(max_count - 5)]
         res = self.app.post_json(self.root + '/storage/col2', bsos)
         res = res.json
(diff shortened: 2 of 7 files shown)


syncstorage/tests/functional/test_old_storage.py
```diff
@@ -412,32 +411,6 @@ def test_get_collection_ttl(self):
...
-        self.assertEquals(len(res['success']), max_count)
-        self.assertEquals(len(res['failed']), 5)
-        # The test config has max_bytes=1M.
-        # Uploading 5 210MB items should produce one failure.
-        wbos = [{'id': str(i), 'payload': "X" * (210 * 1024)}
-                for i in range(5)]
-        res = self.app.post_json(self.root + '/storage/col2', wbos)
-        res = res.json
-        self.assertEquals(len(res['success']), 4)
-        self.assertEquals(len(res['failed']), 1)
```
rfk: These tests just assert compatibility with the old sync1.1 API, which we no longer care about, so I opted to delete this test rather than fix it.

syncstorage/tests/functional/test_storage.py
```diff
@@ -780,15 +780,20 @@ def test_get_collection_ttl(self):
         self.assertEquals(len(res.json), 0)
 
     def test_batch(self):
-        # This can't be run against a live server
```
rfk: It now *can* be run against a live server, if it returns the proper config info \o/

syncstorage/views/__init__.py
```diff
@@ -213,12 +215,11 @@ def get_info_configuration(request):
     else:
         LIMIT_NAMES = (
             "max_request_bytes",
+            "max_record_payload_bytes",
         )
     limits = {}
     for name in LIMIT_NAMES:
         limits[name] = get_limit_config(request, name)
-    # This limit is hard-coded for now.
-    limits["max_record_payload_bytes"] = MAX_PAYLOAD_SIZE
```
rfk: I moved this under the control of `get_limit_config` so that it can be adjusted via config file, which is important for the memcached tests.

syncstorage/views/validators.py
```diff
@@ -303,12 +304,13 @@ def parse_multiple_bsos(request):
             logger.info(logmsg, userid, collection, id, msg, bso)
             continue
 
-        if count >= BATCH_MAX_COUNT:
```
rfk: This was accidentally an inclusive range, because `enumerate` was starting the count at 0.

syncstorage/tests/functional/test_storage.py
```diff
@@ -802,15 +807,16 @@ def test_batch(self):
...
-        # The test config has max_bytes=1M.
-        # Uploading 5 210MB items should produce one failure.
-        max_bytes = get_limit_config(self.config, 'max_post_bytes')
-        self.assertEquals(max_bytes, 1024 * 1024)
-        bsos = [{'id': str(i), 'payload': "X" * (210 * 1024)}
-                for i in range(5)]
+        # Uploading N+1 210kB items should produce one failure.
+        item_size = (210 * 1024)
+        max_items = max_bytes / item_size
+        self.assertTrue(max_items * item_size < max_bytes)
```
mhammond: this seems to just be checking |(max_bytes / item_size) * item_size < max_bytes| (and similarly below)

syncstorage/views/util.py
```diff
@@ -89,12 +90,19 @@ def get_resource_timestamp(request):
...
 DEFAULT_LIMITS = {}
-DEFAULT_LIMITS["max_request_bytes"] = 1024 * 1024
+DEFAULT_LIMITS["max_record_payload_bytes"] = MAX_PAYLOAD_SIZE
 DEFAULT_LIMITS["max_post_records"] = 100
-DEFAULT_LIMITS["max_post_bytes"] = DEFAULT_LIMITS["max_request_bytes"]
+DEFAULT_LIMITS["max_post_bytes"] = 2 * 1024 * 1024
 DEFAULT_LIMITS["max_total_records"] = 100 * DEFAULT_LIMITS["max_post_records"]
 DEFAULT_LIMITS["max_total_bytes"] = 100 * DEFAULT_LIMITS["max_post_bytes"]
 
+# In production, the request-size limit is actually controlled by nginx.
```
mhammond: I got confused in IRC about this and still am :) Is nginx's size only talking about payloads too?  (Note I've no reason to believe this is wrong, I'm just confused :)

syncstorage/views/validators.py
thomcc: To be clear, it's still an inclusive range now, it's just more explicitly inclusive no?

syncstorage/views/util.py
thomcc: I'm pretty sure nginx's size is concerned with total request size, and not payloads.

syncstorage/tests/functional/test_storage.py
```diff
@@ -1509,24 +1515,87 @@ def test_batch_size_limits(self):
...
 
+        bsos = [{'id': str(x), 'payload': ''}
+                for x in xrange(limits['max_post_records'])]
+        res = self.app.post_json(endpoint, bsos)
+        self.assertFalse(res.json['failed'])
+        bsos.append({'id': 'toomany', 'payload': ''})
+        res = self.app.post_json(endpoint, bsos)
+        self.assertEquals(res.json['failed']['toomany'], 'retry bso')
+
+        # `max_total_records` is an (inclusive) limit on the
```
thomcc: I don't think desktop will say this. Maybe depend is the wrong word then? (Or we need a client fix...). Same for `X-Weave-Total-Bytes` below.

syncstorage/tests/functional/test_storage.py
```diff
@@ -802,15 +807,16 @@ def test_batch(self):
...
+        max_items = max_bytes / item_size
+        self.assertTrue(max_items * item_size < max_bytes)
+        self.assertTrue((max_items + 1) * item_size > max_bytes)
+        bsos = [{'id': str(i), 'payload': "X" * item_size}
+                for i in range(max_items + 1)]
         res = self.app.post_json(self.root + '/storage/col2', bsos)
         res = res.json
-        self.assertEquals(len(res['success']), 4)
+        self.assertEquals(len(res['success']), max_items)
         self.assertEquals(len(res['failed']), 1)
```
thomcc: Arguably, everything should fail inside a batch if a single record fails... (Normally the client could just not commit the batch, but that's not an option for batch=1234&commit=true where the body has records, which for desktop – and probably the other clients – it almost always will). But that's certainly a discussion for a different time...

I guess if we respect all the config limits, there should be no failures, so it probably doesn't matter.

syncstorage/tests/functional/test_storage.py
```diff
@@ -802,15 +807,16 @@ def test_batch(self):
         self.assertEquals(len(res['success']), max_count)
         self.assertEquals(len(res['failed']), 5)
 
-        # The test config has max_bytes=1M.
-        # Uploading 5 210MB items should produce one failure.
-        max_bytes = get_limit_config(self.config, 'max_post_bytes')
-        self.assertEquals(max_bytes, 1024 * 1024)
-        bsos = [{'id': str(i), 'payload': "X" * (210 * 1024)}
-                for i in range(5)]
+        # Uploading N+1 210kB items should produce one failure.
```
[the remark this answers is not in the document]
rfk: You understand correctly; I'll adjust the comment to make it more obvious what this is about.

syncstorage/tests/functional/test_storage.py
rfk: Yes, the idea was to check that `item_size` is not an exact multiple of `max_bytes` but I don't think that's actually necessary, I'll remove it.

syncstorage/tests/functional/test_storage.py
rfk: Yeah, I agree, and in fact I feel like it would be simpler all round if any bad item in a request just failed the whole request.  But I think we kept it this way just so we didn't have to mess with existing client code, and particularly so we didn't have different behaviour between batch and non-batch cases.

syncstorage/tests/functional/test_storage.py
rfk: "We can only enforce it if..." is probably a more accurate phrasing.  Of course iif desktop *can* send this it would be helpful :-)

syncstorage/views/util.py
rfk: Correct, nginx will basically reject any request with content-length over its configured limit.

syncstorage/views/validators.py
rfk: Correct, it's now more obvious that it's an inclusive range on purpose.

PR #75, opened 2017-10-03
Bug 1388963 - Emit timing info for batch purge operations.

rfk:


rfk:
I'm going to continue this over in #76

PR #76, opened 2017-10-04
Bug 1388963 - Run `OPTIMIZE TABLE` after purging items

rfk:
WIP, opening for CI

### DIFF
--- a/syncstorage/storage/sql/__init__.py
+++ b/syncstorage/storage/sql/__init__.py
@@ -804,7 +804,7 @@ class SQLStorage(SyncStorage):
         num_purged = 0
         is_incomplete = False
         # Note that we take a new session for each run of the query.
-        # This avoids holdig open a long-running transaction, so
+        # This avoids holding open a long-running transaction, so
         # the incrementality can let other jobs run properly.
         with self._get_or_create_session() as session:
             params["now"] = int(session.timestamp)
--- a/syncstorage/storage/sql/__init__.py
+++ b/syncstorage/storage/sql/__init__.py
@@ -128,6 +128,8 @@ class SQLStorage(SyncStorage):
 
         self.sqluri = sqluri
         self.dbconnector = DBConnector(sqluri, **dbkwds)
+        self._optimize_table_after_purge = \
+            dbkwds.get("optimize_table_after_purge", True)
 
         # There doesn't seem to be a reliable cross-database way to set the
         # initial value of an autoincrement column.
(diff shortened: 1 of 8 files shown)


syncstorage/storage/sql/__init__.py
```diff
@@ -821,6 +823,18 @@ def _purge_items_loop(self, table, query, params):
...
+        #
+        #   https://www.percona.com/blog/2015/02/11/tokudb-table-optimization-improvements/
+        #
+        # We run an `OPTIMIZE TABLE` after each purge to keep this in check.
+        optimize_frac = self._tokudb_optimize_index_faction
+        if optimize_frac and self.dbconnector.driver == "mysql":
+            q = "SET tokudb_optimize_index_fraction = %.2f" % (optimize_frac,)
+            with self._get_or_create_session() as session:
+                session.query(q)
+                session.query("OPTIMIZ TABLE %s" % (table,))
```
rfk: Deliberate typo so I can see if CI will actually cause this line to execute :-)

rfk:
OK, I think this is ready to go.  There are two commits here.

The first, courtesy of the "write a failing test case first" department, fixes .travis.yml so that we actually run the tests on mysql and postgres backends in CI.  It also fixes some edge-case bugs uncovered by making the tests actually run.

The second is the core of this PR.  It adds a call to `OPTIMIZE TABLE` after doing the background prune operation, adds a config flag to let you pref it on or off, adds timers so we can monitor whether perf is getting out of hand on these tables, and fiddles with the test config to ensure that the code actually gets run.

syncstorage/storage/sql/__init__.py
```diff
@@ -766,6 +769,16 @@ def _purge_expired_batches(self, grace_period=0, max_per_loop=1000):
...
             "maxitems": max_per_loop,
         })
+        # For TokuDB tables, doing a bunch of in-order deletes can cause
+        # performance degradation of the indexes, as described in:
+        #
+        #   https://www.percona.com/blog/2015/02/11/tokudb-table-optimization-improvements/
+        #
+        # We run an `OPTIMIZE TABLE` after each purge to keep this in check.
+        if self._optimize_table_after_purge:
+            if self.dbconnector.driver == "mysql":
```
kitcambridge: What do you think about moving these conditions into a Boolean getter on `SQLStorage`, so that you don't need to duplicate the comment?

syncstorage/storage/sql/dbconnect.py
```diff
@@ -108,7 +108,9 @@ def __new__(cls, *args, **kwds):
 # It is used to create either sharded or non-shareded BSO storage tables,
 # depending on the run-time settings of the application.
 
-PAYLOAD_TYPE = Text().with_variant(postgresql.TEXT(), 'postgresql')
+PAYLOAD_TYPE = Text()
+PAYLOAD_TYPE = PAYLOAD_TYPE.with_variant(postgresql.TEXT(), 'postgresql')
+PAYLOAD_TYPE = PAYLOAD_TYPE.with_variant(mysql.MEDIUMTEXT(), 'mysql')
```
kitcambridge: Clever!

rfk:
Thanks @kitcambridge, merging with your suggested cleanup

PR #77, opened 2017-12-07
Run `OPTIMIZE TABLE` before a purge, as well as after.

rfk:
This will help ensure that the optimization actually gets run.
Without it, a highly-garbage-filled index could cause the purge
job to take a long time and never get around to actually running
the optimize.

### DIFF
--- a/syncstorage/storage/sql/__init__.py
+++ b/syncstorage/storage/sql/__init__.py
@@ -128,6 +128,8 @@ class SQLStorage(SyncStorage):
 
         self.sqluri = sqluri
         self.dbconnector = DBConnector(sqluri, **dbkwds)
+        self._optimize_table_before_purge = \
+            dbkwds.get("optimize_table_before_purge", True)
         self._optimize_table_after_purge = \
             dbkwds.get("optimize_table_after_purge", True)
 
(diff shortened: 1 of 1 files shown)


syncstorage/storage/sql/__init__.py
```diff
@@ -764,12 +766,13 @@ def _purge_expired_bsos(self, grace_period=0, max_per_loop=1000):
...
 
     def _purge_expired_batches(self, grace_period=0, max_per_loop=1000):
+        self._maybe_optimize_table_before_purge("OPTIMIZE_BATCHES_TABLE")
         return self._purge_items_loop("batch_uploads", "PURGE_BATCHES", {
             "lifetime": BATCH_LIFETIME,
             "grace": grace_period,
             "maxitems": max_per_loop,
         })
-        self._maybe_optimize_table("OPTIMIZE_BATCHES_TABLE")
+        self._maybe_optimize_table_after_purge("OPTIMIZE_BATCHES_TABLE")
```
kitcambridge: Unreachable code after `return`? (I think I missed this when you originally added `_maybe_optimize_table`, apologies!)

rfk: Nice catch!  I'm astonished this didn't fail linting, digging in now to see if there's a problem with the linter...

syncstorage/storage/sql/__init__.py
```diff
@@ -792,7 +798,9 @@ def _purge_expired_batch_items(self, grace_period=0, max_per_loop=1000):
             })
             num_purged += res["num_purged"]
             is_incomplete = is_incomplete or not res["is_complete"]
-        self._maybe_optimize_table("OPTIMIZE_BUI_TABLE", {"bui": table})
+            self._maybe_optimize_table_after_purge("OPTIMIZE_BUI_TABLE", {
+                "bui": table
+            })
```
rfk: Also there was a bug with the indentation here; previously it would only run after the entire loop had completed, optimizing only the last table in the list instead of optimizing each table in turn :-(

syncstorage/storage/sql/__init__.py
rfk: Looks like flake8 doesn't catch this as an error, but pylint does.  There's a separate project going on to migrate some of these repos over to use pylint, so I'll just fix the error here and leave the broader linting question alone for now...

rfk:
Thanks @kitcambridge, updated with a fix to make the code actually run!

PR #78, opened 2018-02-07
Avoid triggering a 413 with large payload tests in prod.

rfk:
This test was wildly overshooting the nginx limit on request body size when run against servers in stage.  I've adjusted it to be a little more conservative about how much data it sends, while still triggering the behaviour under test (which is the max BSO payload bytes, not max request bytes).

@Micheletto r?

### DIFF
--- a/syncstorage/tests/functional/test_storage.py
+++ b/syncstorage/tests/functional/test_storage.py
@@ -18,6 +18,7 @@ import unittest2
 
 import re
 import sys
+import json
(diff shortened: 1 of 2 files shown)


Micheletto:
> @Micheletto r?

r+

rfk:
@Micheletto thanks!  I made a 1.6.13 tag with this fix included if you want to use it for testing stage.

PR #81, opened 2018-05-02
Migrate to CircleCI 2.0

autrilla:
Build works on my fork (https://circleci.com/gh/autrilla/server-syncstorage/4)


autrilla:
The docker version could be abstracted with an env var, but i don't think that's doable for the ubuntu image version, if that's what you mean

autrilla:
I've parametrized the docker version, `DOCKER_VERSION` will have to be set to `17.12.1-ce` on CircleCI

.circleci/config.yml
```diff
@@ -0,0 +1,72 @@
...
+            "$CIRCLE_PROJECT_REPONAME" \
+            "$CIRCLE_BUILD_URL" > version.json
+      - store_artifacts:
+          path: version.json
+      # FIXME: should use an image w/ docker installed by default
+      - run:
+          name: Install Docker
+          command: |
+            set -x
+            curl -L -o /tmp/docker-$DOCKER_VERSION.tgz https://download.docker.com/linux/static/stable/x86_64/docker-$DOCKER_VERSION.tgz
```
rfk: Is there any way we can integrity-check the result of this download, even if it's just comparing its `sha256sum` against a known good value?

I guess the app itself doesn't integrity-check its dependencies either, but it probably should...

autrilla: I've added a check for the sha256 of the download

autrilla:
@rfk can this be merged?

rfk:
👍

PR #85, opened 2018-05-30
Bug 1464008 - explicitly order by BSO id.

rfk:
It turns out we can't assume that the DB will return results in a consistent order on newer versions of MysQL, so we'll have to do it explicitly.  We will need to do some query-explaining to ensure that MySQL doesn't turn this into a filesort.

ref https://bugzilla.mozilla.org/show_bug.cgi?id=1464008

### DIFF
--- a/syncstorage/storage/sql/queries_generic.py
+++ b/syncstorage/storage/sql/queries_generic.py
@@ -223,16 +223,15 @@ def FIND_ITEMS(bso, params):
     # Sort it in the order requested.
     # We always sort by *something*, so that limit/offset work consistently.
     # The default order is by timestamp, which if efficient due to the index.
-    # NOTE: ideally we would sort by "id" here as secondary column, to get a
-    # consistent total ordering.  But we don't want to bloat the index, so
-    # we just assume that the db gives results in a consistent order.
+    # We sort by "id" as a secondary column to ensure a consistent total
+    # ordering, which is important when paginating large requests.
     sort = params.get("sort", None)
     if sort == 'index':
-        query = query.order_by(bso.c.sortindex.desc())
+        query = query.order_by(bso.c.sortindex.desc(), bso.c.id.desc())
     elif sort == 'oldest':
-        query = query.order_by(bso.c.modified.asc())
+        query = query.order_by(bso.c.modified.asc(), bso.c.id.asc())
     else:
-        query = query.order_by(bso.c.modified.desc())
+        query = query.order_by(bso.c.modified.desc(), bso.c.id.asc())
     # Apply limit and/or offset.
     limit = params.get("limit", None)
     if limit is not None:


autrilla:
It looks like it works!

I have done the following:

- Disabled verification in Jenkins
- Deployed a storage node using the latest CircleCI image
- On the storage node:
  - Tagged the CircleCI image as `sync-original`
  - Copied over the `queries_generic.py` file.
  - Built an image with the name and tag the CircleCI image used, effectively replacing it. With the following dockerfile:
    ```
    FROM sync-original
    COPY queries_generic.py /app/syncstorage/storage/sql/
    ```
  - Restarted the docker container so that it uses the new image
- Then, on the node manager:
  - Ran the verification script (wrapped in a Python helper).
  - The results: no failures, 30 successes straight 🎉 

@rfk what queries do you want `EXPLAIN`ed?

Do we want to merge this to master and load test it with all the changes that will happen with the changes I'm making, or do we want this in a separate branch (or gated through a feature flag) to load test it separately?

Micheletto:
> Do we want to merge this to master and load test it with all the changes that will happen with the changes I'm making, or do we want this in a separate branch (or gated through a feature flag) to load test it separately?

I think our best understanding of the performance implications of this change would be to do some before and after testing on the old style SL6 nodes.  There's already so many changing variables with these new type of nodes that it would be difficult to know the implications.  The MySQL slow log dashboard should prove to be useful.

autrilla:
Sounds good. The implications might be totally different with the old MySQL vs the new one, so we have to be aware that the results might not translate well.

In any case, AFAICT we _must_ sort by id for the new MySQL to work correctly, so there isn't much of a point in benchmarking the new nodes with/without this patch unless we find another workaround.

rfk:
> It looks like it works!

Awesome!

> @rfk what queries do you want EXPLAINed?

Something like the queries from https://bugzilla.mozilla.org/show_bug.cgi?id=1025735#c18, which would be:

```
EXPLAIN SELECT bso.userid, bso.collection, bso.id, bso.sortindex, bso.modified, bso.payload, bso.payload_size, bso.ttl  FROM bso  WHERE bso.userid = XXXXXXX AND bso.collection = 4 AND bso.modified > 1403218499710 AND bso.ttl > 1403545706 ORDER BY bso.modified DESC, bso.id DESC LIMIT 5001;
```

and

```
EXPLAIN SELECT bso.userid, bso.collection, bso.id, bso.sortindex, bso.modified, bso.payload, bso.payload_size, bso.ttl  FROM bso  WHERE bso.userid = XXXXXXX AND bso.collection = 4 AND bso.modified > 1403218499710 AND bso.ttl > 1403545706 ORDER BY bso.modified ASC, bso.id ASC LIMIT 5001;
```

We want to check whether they're "Using filesort" or whether MySQL is smart enough to answer them using just the existing indexes for sorting.

I'm not sure whether the query planner will make different choices based on how full the database is, but it may be worth running a loadtest on one of these nodes before hand, to fill up the db with some plausible-looking data.  @Micheletto should be able to help with the details of running such.

rfk:
> In any case, AFAICT we must sort by id for the new MySQL to work correctly, so there isn't much of a 
> point in benchmarking the new nodes with/without this patch unless we find another workaround.

Agreed.  We're not expicitly sorting by id right now, precisely because the old MySQL nodes threatened to make it a filesort rather than use an index.  If these new nodes do the same, we'll have to actually add the id column to the index explicitly.

autrilla:
The `bso` table didn't exist, I used `bso0`:

```
MariaDB [weave0]> EXPLAIN SELECT bso.userid, bso.collection, bso.id, bso.sortindex, bso.modified, bso.payload, bso.payload_size, bso.ttl  FROM bso0 AS bso  WHERE bso.userid = 660 AND bso.collection = 4 AND bso.modified > 1403218499710 AND bso.ttl > 1403545706 ORDER BY bso.modified DESC, bso.id DESC LIMIT 5001;
+------+-------------+-------+-------+-----------------------------------------+---------------------+---------+------+------+-------------+
| id   | select_type | table | type  | possible_keys                           | key                 | key_len | ref  | rows | Extra       |
+------+-------------+-------+-------+-----------------------------------------+---------------------+---------+------+------+-------------+
|    1 | SIMPLE      | bso   | range | PRIMARY,bso_ttl_idx,bso_usr_col_mod_idx | bso_usr_col_mod_idx | 16      | NULL |    1 | Using where |
+------+-------------+-------+-------+-----------------------------------------+---------------------+---------+------+------+-------------+
1 row in set (0.00 sec)
```

And the other one:

```
MariaDB [weave0]> EXPLAIN SELECT bso.userid, bso.collection, bso.id, bso.sortindex, bso.modified, bso.payload, bso.payload_size, bso.ttl  FROM bso0 AS bso  WHERE bso.userid = 660 AND bso.collection = 4 AND bso.modified > 1403218499710 AND bso.ttl > 1403545706 ORDER BY bso.modified ASC, bso.id ASC LIMIT 5001;
+------+-------------+-------+-------+-----------------------------------------+---------------------+---------+------+------+-------------+
| id   | select_type | table | type  | possible_keys                           | key                 | key_len | ref  | rows | Extra       |
+------+-------------+-------+-------+-----------------------------------------+---------------------+---------+------+------+-------------+
|    1 | SIMPLE      | bso   | range | PRIMARY,bso_ttl_idx,bso_usr_col_mod_idx | bso_usr_col_mod_idx | 16      | NULL |    1 | Using where |
+------+-------------+-------+-------+-----------------------------------------+---------------------+---------+------+------+-------------+
1 row in set (0.00 sec)
```

So it looks like there's no filesort. No load test has been performed on this node, though.

rfk:
Thanks @autrilla, that looks like good news!  Unfortunately I don't think I'll have time to push on this any further until after the all-hands.  We need to figure out how to make that chance on the new nodes in a way that doesn't affect the old nodes.  Can we please try the `EXPLAIN`s above on an old-style node to see whether they do indeed show a different query plan?

autrilla:
This can certainly wait until after the All Hands, indeed I have things I'd do instead of this before the All Hands! The explains on an old-style node:

```
MariaDB [weave0]> EXPLAIN SELECT bso.userid, bso.collection, bso.id, bso.sortindex, bso.modified, bso.payload, bso.payload_size, bso.ttl  FROM bso0 AS bso  WHERE bso.userid = 660 AND bso.collection = 4 AND bso.modified > 1403218499710 AND bso.ttl > 1403545706 ORDER BY bso.modified DESC, bso.id DESC LIMIT 5001;
+------+-------------+-------+------+-----------------------------------------+---------+---------+-------------+------+-----------------------------+
| id   | select_type | table | type | possible_keys                           | key     | key_len | ref         | rows | Extra                       |
+------+-------------+-------+------+-----------------------------------------+---------+---------+-------------+------+-----------------------------+
|    1 | SIMPLE      | bso   | ref  | PRIMARY,bso_ttl_idx,bso_usr_col_mod_idx | PRIMARY | 8       | const,const |    1 | Using where; Using filesort |
+------+-------------+-------+------+-----------------------------------------+---------+---------+-------------+------+-----------------------------+
1 row in set (0.00 sec)

MariaDB [weave0]> EXPLAIN SELECT bso.userid, bso.collection, bso.id, bso.sortindex, bso.modified, bso.payload, bso.payload_size, bso.ttl  FROM bso0 AS bso  WHERE bso.userid = 660 AND bso.collection = 4 AND bso.modified > 1403218499710 AND bso.ttl > 1403545706 ORDER BY bso.modified ASC, bso.id ASC LIMIT 5001;
+------+-------------+-------+------+-----------------------------------------+---------+---------+-------------+------+-----------------------------+
| id   | select_type | table | type | possible_keys                           | key     | key_len | ref         | rows | Extra                       |
+------+-------------+-------+------+-----------------------------------------+---------+---------+-------------+------+-----------------------------+
|    1 | SIMPLE      | bso   | ref  | PRIMARY,bso_ttl_idx,bso_usr_col_mod_idx | PRIMARY | 8       | const,const |    1 | Using where; Using filesort |
+------+-------------+-------+------+-----------------------------------------+---------+---------+-------------+------+-----------------------------+
1 row in set (0.00 sec)
```

So this confirms we need to ensure this change doesn't go live on the old nodes.

rfk:
Closing in favour of https://github.com/mozilla-services/server-syncstorage/pull/88

PR #86, opened 2018-06-29
Bug 1464008 explicitly order by id config flag

autrilla:
Based on #85 

There's probably a less painful way to pass a config flag all the way down to where the query is generated...

### DIFF
--- a/example.ini
+++ b/example.ini
@@ -50,7 +50,7 @@ pool_recycle = 3600
 reset_on_return = true
 create_tables = true
 batch_max_count = 4000
-
+sort_by_id = true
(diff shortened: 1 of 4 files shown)


rfk:
Thanks for pushing this forward @autrilla!

> There's probably a less painful way to pass a config flag all the way down to where the query is generated

I had a go at simplifying this over in https://github.com/mozilla-services/server-syncstorage/pull/88, let me know what you think.

rfk:
Closing in favour of https://github.com/mozilla-services/server-syncstorage/pull/88

PR #87, opened 2018-07-16
Update gunicorn dependency.

rfk:
This resolves the dependency warning for older versions of gunicorn, per https://github.com/mozilla-services/server-syncstorage/network/dependencies

I don't have any reason to believe there's an actual security issue here, but updating for completeness.


rfk:
Going to need a review on this after enabling protection for master branch...thanks @bbangert :-)

PR #90, opened 2018-09-28
Accept FxA user-id and key-id details in Hawk token.

rfk:
This is a sketch of what might need to change in the storage code to support https://github.com/mozilla-services/tokenserver/issues/121, receiving the raw FxA uid and key-id as part of the tokenserver token.  There would need to be some additional figuring-out of how to pass these through to the storage backend, which currently expects to receive the numeric tokenserver userid as first argument.  Maybe we change it to pass in `request.user` instead of the raw id?

Connects to https://github.com/mozilla-services/tokenserver/issues/121; Depends on  #93.

/cc @bbangert @philbooth @pjenvey

### DIFF
--- a/syncstorage/views/authentication.py
+++ b/syncstorage/views/authentication.py
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import re
 import time
 import logging
 
@@ -17,6 +18,11 @@ logger = logging.getLogger(__name__)
 
 
 DEFAULT_EXPIRED_TOKEN_TIMEOUT = 60 * 60 * 2  # 2 hours, in seconds
+
+# Coarse validation of FxA userid, device ids, and key ids.
+# This is not supposed to catch all invalid cases, but to act as a backstop
+# that the ids are safe to use and store internally.
+VALID_FXA_ID_REGEX = re.compile("^[A-Za-z0-9=-]{1,64}$")
 
 
 class SyncStorageAuthenticationPolicy(TokenServerAuthenticationPolicy):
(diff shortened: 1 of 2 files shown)


rfk:
I've rebased this to depend on the big refactor in https://github.com/mozilla-services/server-syncstorage/pull/93; that PR should be reviewed and merged first, and then this PR can be edited to target `master` branch.

rfk:
Also, it's **NOT SAFE TO MERGE** this until we've resolved open questions around changing the name of keys in `request.metrics`.

rfk:
Now with tests!

rfk:
I decided to add some additional validation to ensure that the "fxa_uid" and "fxa_kid" are well-formed, since they originate outside the system.  @bbangert could you please take another look before merge?

syncstorage/views/authentication.py
```diff
@@ -70,46 +76,142 @@ def decode_hawk_id(self, request, tokenid):
...
+        # Sanity-check that we're on the right node.
+        if data["node"] != node_name:
             msg = "incorrect node for this token: %s"
-            raise ValueError(msg % (token_node_name,))
+            raise ValueError(msg % (data["node"],))
+
         # Calculate the matching request-signing secret.
         key = tokenlib.get_derived_secret(tokenid, secret=secret)
 
-        request.metrics["fxa_uid"] = data.get("fxa_uid")
```
bbangert: I assume backend systems relying on the id's are updated somewhere?

rfk: We could only find the one place that this is being used, and it doesn't sound like anyone's using it; we'll either update or disable it in https://bugzilla.mozilla.org/show_bug.cgi?id=1500830

syncstorage/views/authentication.py
```diff
@@ -70,46 +76,142 @@ def decode_hawk_id(self, request, tokenid):
...
+            user["hashed_fxa_uid"] = data["hashed_fxa_uid"]
+            if not VALID_FXA_ID_REGEX.match(user["hashed_fxa_uid"]):
+                raise ValueError("invalid hashed_fxa_uid in token data")
+            try:
+                user["fxa_uid"] = data["fxa_uid"]
+            except KeyError:
+                raise ValueError("missing fxa_uid in token data")
+            else:
+                if not VALID_FXA_ID_REGEX.match(user["fxa_uid"]):
+                    raise ValueError("invalid fxa_kid in token data")
```
pjenvey:
```suggestion
                    raise ValueError("invalid fxa_uid in token data")
```

syncstorage/views/authentication.py
```diff
@@ -70,46 +76,142 @@ def decode_hawk_id(self, request, tokenid):
...
+                if not VALID_FXA_ID_REGEX.match(user["fxa_kid"]):
+                    raise ValueError("invalid fxa_kid in token data")
+        elif "fxa_uid" in data:
+            user["hashed_fxa_uid"] = data["fxa_uid"]
+            if not VALID_FXA_ID_REGEX.match(user["hashed_fxa_uid"]):
+                raise ValueError("invalid fxa_uid in token data")
+
+        if "hashed_device_id" in data:
+            user["hashed_device_id"] = data["hashed_device_id"]
+            if not VALID_FXA_ID_REGEX.match(user["hashed_fxa_uid"]):
```
pjenvey:
```suggestion
            if not VALID_FXA_ID_REGEX.match(user["hashed_device_id"]):
```

syncstorage/views/authentication.py
```diff
@@ -70,46 +76,142 @@ def decode_hawk_id(self, request, tokenid):
...
+                raise ValueError("invalid hashed_device_id in token data")
+        elif "device_id" in data:
+            user["hashed_device_id"] = data.get("device_id")
+            if not VALID_FXA_ID_REGEX.match(user["hashed_fxa_uid"]):
```
pjenvey:
```suggestion
            if not VALID_FXA_ID_REGEX.match(user["hashed_device_id"]):
```

syncstorage/views/authentication.py
```diff
@@ -18,6 +19,11 @@
 
 DEFAULT_EXPIRED_TOKEN_TIMEOUT = 60 * 60 * 2  # 2 hours, in seconds
 
+# Coarse validation of FxA userid, device ids, and key ids.
+# This is not supposed to catch all invalid cases, but to act as a backstop
+# that the ids are safe to use and store internally.
+VALID_FXA_ID_REGEX = re.compile("^[A-Za-z0-9=-]{1,64}$")  #
```
pjenvey: nit, this looks stray:
```suggestion
VALID_FXA_ID_REGEX = re.compile("^[A-Za-z0-9=-]{1,64}$")
```

rfk:
Thanks @pjenvey! I accepted the suggestions and squashed into a single commit for clarity.

PR #91, opened 2018-10-06
Change installation of umemcache

okin:
`umemcache` is an optional (extra) dependency of `mozsvc`.
`umemcache` is never used directly in this module.

Since `umemcache` is used with the version of `mozsvc.storage.memcached` installed the installation style for `mozsvc` has been changed to include the extra `memcache` - which will currently install `umemcache`.
If `mozsvc` changes its dependencies for _memcache_ these changes can be picked up much easier with this style.

My main motivation behind this is that porting this library to Python 3 gets easier without `umemcache` as a direct dependency because it only supports Python 2.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,13 +3,11 @@
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
 from setuptools import setup, find_packages
-import os
-import re
 
 install_requires = ['SQLALchemy', 'unittest2', 'simplejson', 'pyramid',
-                    'mozsvc>=0.8', 'cornice', 'pyramid_hawkauth', 'PyMySQL',
-                    'pymysql_sa', 'umemcache', 'wsgiproxy',
-                    'webtest', 'requests', 'PyBrowserID', 'testfixtures']
+                    'mozsvc[memcache]>=0.8', 'cornice', 'pyramid_hawkauth',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -3,13 +3,11 @@
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
 from setuptools import setup, find_packages
-import os
-import re
 
 install_requires = ['SQLALchemy', 'unittest2', 'simplejson', 'pyramid',
                     'mozsvc>=0.8', 'cornice', 'pyramid_hawkauth', 'PyMySQL',
```
rfk: Should the `mozsvc` dependency here be changed similarly?  It's been a while since I did anything with python dependency management so I'm a bit rusty on the details.

okin: I wasn't sure so I ran some tests (Debian 9, Python 2.7, pip 18.1).

I added `[memcache]` to the _mozsvc_ dependency and tried installing.
If you install with pip (e.g. `pip install .` from inside the working copy) then the additional package will be picked up as expected.
If you use `python setup.py install` then _umemcache_ will still be installed but in my case I ended up running into problems with compilation of _umemcache_ - bot in my branch and _master_.

I will adjust the dependency here accordingly.

okin:
Sure, no problem.
I'll try to keep the squashing in mind for future pull requests.

PR #92, opened 2018-10-06
Travis improvements

okin:
Travis CI is a great tool for having your package tested.

This pull requests:
* Drops support for Python 2.6 (seems to be out-of-scope with deps not even running on this old Python; as per [this comment](https://github.com/mozilla-services/tokenserver/issues/102#issuecomment-281698235) production runs Python 2.7)
* Switches pyscopg2 to psycobg2-binary, which is a prepared wheel that will cut down installation time
* Will let the makefile create a virtualenv with the systems Python. This gets important if you want to run multiple Python versions.
* Updates the URL of the PyPI index to the one used by pypi.org. The old one does not receive updates.
* Fixes building umemcache (was broken once the switch to system python was made)
* Switches from Distribute to setuptools because nowadays the code bases have been merged.
* Updates the used flake8 (and adds an workaround for a problem with older versions of pycodestyle).

This should benefit everyone who wants to work on this with the help of Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 
 python:
-  - "2.6"
   - "2.7"
 
 services:
@@ -20,7 +19,7 @@ notifications:
 
 install:
   - make build
-  - ./local/bin/pip install psycopg2
+  - ./local/bin/pip install psycopg2-binary
 
 before_script:
   - mysql -e 'create database sync_test;'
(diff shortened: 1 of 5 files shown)


rfk:
Thanks @okin!  Could you please also add yourself to `CONTRIBUTORS.txt` for completeness?  Also if you're happy to squash these into a single commit that would be helpful, but not necessary.

okin:
@rfk added myself to the contributors and squashed.

I'm happy to help :)

PR #93, opened 2018-10-09
feat(backend): Pass full user object to backend storage implementations.

rfk:
This passes the full `request.user` object through to backend storage implementations, rather than just the numeric userid.  With this change in place, a future backend could choose to grab the `fxa_uid` and `fxa_kid` keys out of `request.user` and use those for indexing into storage, rather than using the numeric id allocated by tokenserver.

It's a big change, but it's largely mechanical, and AFAIK it should be entirely contained to this repo.  The only weird bit was the handling of expired tokens, which took me a few attempts to get figured out, but I think has successfully maintained its simplicity and safety.

@bbangert thoughts? (and if it seems like a reasonable approach, r?)

Connects to https://github.com/mozilla-services/tokenserver/issues/121

### DIFF
--- a/syncstorage/scripts/mcclear.py
+++ b/syncstorage/scripts/mcclear.py
@@ -48,7 +48,7 @@ def clear_memcache_data(config_file, input_file):
             uid = uid.strip()
             if uid:
                 logger.info("Clearing data for %s", uid)
-                for key in backend.iter_cache_keys(uid):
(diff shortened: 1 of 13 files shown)


rfk:
The test covering the expired-token stuff may also provide useful context: https://github.com/mozilla-services/server-syncstorage/blob/master/syncstorage/tests/functional/test_storage.py#L1307

PR #94, opened 2018-11-01
Update requests version dependency to silence security alert.

rfk:


pjenvey:
I don't have the perms to merge this one myself

PR #95, opened 2018-12-20
feat: add google spanner storage back-end

bbangert:
Co-authored by Phil Jenvey <pjenvey@underboss.org>

Add's Google Cloud Spanner storage back-end, with minor tweaks to tween
for X-W-TS handling.


syncstorage/storage/spanner.py
```diff
@@ -0,0 +1,1054 @@
...
+                self._collections_by_id[id] = name
+
+        # A thread-local to track active sessions.
+        self._tldata = threading.local()
+
+    def _lock_write_precondition(self, session, user, collection):
+        """'Lock' the precondition on the collection
+
+        Get the timestamp of the collection in use, utilizing it as precondition for
+        transaction invalidation.
```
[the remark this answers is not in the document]
bbangert: Yup, or if no row was present here, then its expected that case will hold during the transaction.

syncstorage/storage/spanner.py
```diff
@@ -0,0 +1,1054 @@
...
+                "SELECT CURRENT_TIMESTAMP()"
+            ).one()[0]
+            session.timestamp = dt2ts(result)
+            session.collection_ts = 0
+
+    def lock_for_read(self, viewfunc, request, user, collection):
+        with SpannerStorageSession(self) as spanner_session:
+            with self._database.snapshot(multi_use=True) as txn:
+                spanner_session.set_transaction(txn)
+                return viewfunc(request)
```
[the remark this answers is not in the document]
bbangert: Yea, apparently there's some conditions in which the grpc layer wants to try it again. We extracted the try logic from the Spanner lib, as the only way they exposed the read-write transaction required passing in the function (and the only return value you get is the final commit timestamp). That obviously wasn't going to work, so we had to special-case the lock to get the whole viewfunc, so the retry for spanner/grpc client issues could be performed.

It's somewhat of an unknown how frequently this actually occurs.

syncstorage/storage/spanner.py
```diff
@@ -0,0 +1,1054 @@
...
+                    txn.begin()
+                    self._lock_write_precondition(spanner_session, user, collectionid)
+                    try:
+                        result = viewfunc(request)
+                    except Aborted as exc:
+                        del session._transaction
+                        # XXX:
+                        #_delay_until_retry(exc, deadline)
+                        if not spanner_session.comitted:
+                            continue
```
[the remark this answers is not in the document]
bbangert: Indeed, this is the extracted internal Spanner retry logic that is used. I believe its to handle other fail conditions in the underlying grpc/spanner protocols that is not a Conflict error.

syncstorage/storage/spanner.py
```diff
@@ -0,0 +1,1054 @@
...
+                         "collectionid": param_types.INT64}
+        ).one_or_none()
+        if result:
+            current_ts, last_modified = result
+            session.collection_ts = dt2ts(last_modified)
+            session.timestamp = dt2ts(current_ts)
+            # Forbid the write if it would not properly incr the timestamp.
+            if session.collection_ts >= dt2ts(current_ts):
+                msg = "last_modified: %s (%s) current_ts: %s (%s)" % (last_modified, session.collection_ts, current_ts, dt2ts(current_ts))
+                raise ConflictError("lock_for_write %s" % msg)
```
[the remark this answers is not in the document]
bbangert: This function is only called in lock_for_write though?

syncstorage/storage/spanner.py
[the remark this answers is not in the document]
bbangert: This has been changed quite a bit since this review.

syncstorage/views/decorators.py
```diff
@@ -189,5 +189,10 @@ def with_collection_lock(viewfunc, request):
         lock_collection = storage.lock_for_read
     else:
         lock_collection = storage.lock_for_write
+
+    # If we're using spanner, pass in the function for transaction begin
+    if storage.sqluri.startswith("spanner"):
+        return lock_collection(viewfunc, request, user, collection)
```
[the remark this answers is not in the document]
bbangert: Outdated.

syncstorage/storage/spanner.py
```diff
@@ -0,0 +1,1054 @@
...
+            )
+        res = self._map_collection_names(res)
+        for collection in res:
+            res[collection] = dt2ts(res[collection])
+        return res
+
+    def get_collection_counts(self, user):
+        """Returns the collection counts."""
+        userid = user["fxa_uid"]
+        with self._database.snapshot() as snapshot:
```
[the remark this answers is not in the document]
bbangert: This is no longer done.

syncstorage/storage/spanner.py
```diff
@@ -0,0 +1,988 @@
...
+        self.client = spanner.Client()
+        self._instance = self.client.instance(instance_id)
+        self._pool = pool = spanner.BurstyPool()
+        self._database = self._instance.database(database_id, pool=pool)
+
+        # There doesn't seem to be a reliable cross-database way to set the
+        # initial value of an autoincrement column.
+        self.standard_collections = standard_collections
+        if self.standard_collections and dbkwds.get("create_tables", False):
+            pass
```
rfk: If this is still WIP, may be worth throwing here or something rather than just `pass`ing?

syncstorage/storage/spanner.py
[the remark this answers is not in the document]
rfk: Fair point; I think I had read the previous version of this as also being called from `lock_for_read`, but could easily have misinterpreted.

syncstorage/storage/spanner.py
```diff
@@ -0,0 +1,988 @@
...
+            result = session.transaction.execute_sql(
+                "SELECT CURRENT_TIMESTAMP()"
+            ).one()[0]
+            session.timestamp = dt2ts(result)
+            session.collection_ts = 0
+
+    @contextlib.contextmanager
+    def lock_for_read(self, ruser, collection):
+        with SpannerStorageSession(self) as spanner_session:
+            with self._database.snapshot(multi_use=True) as txn:
```
rfk: Snapshots feel nicer for this than a "read lock" 👍

syncstorage/storage/spanner.py
```diff
@@ -0,0 +1,988 @@
...
+    @with_session
+    def get_collection_sizes(self, session, user):
+        """Returns the total size for each collection."""
+        userid = user_key(user)
+        res = session.transaction.execute_sql(
+            COLLECTIONS_SIZES,
+            params={"userid": userid},
+            param_types={"userid": param_types.STRING}
+        )
+        # Some db backends return a Decimal() instance for this aggregate.
```
rfk: Does "some db backends" still apply in this case where you have a single known db backend?

syncstorage/storage/spanner.py
```diff
@@ -0,0 +1,988 @@
...
+    #
+
+    @with_session
+    def get_collection_timestamp(self, session, user, collection):
+        """Returns the last-modified timestamp of a collection."""
+        userid = user_key(user)
+        collectionid = self._get_collection_id(collection)
+        # The last-modified timestamp may be cached on the session.
+        if session.collection_ts:
+            return session.collection_ts
```
rfk: Does this assume that calling code will only ever access a single collection within the scope of a session?  I wonder if it's worth asserting that explicitly somehow.

syncstorage/storage/spanner.py
```diff
@@ -0,0 +1,988 @@
...
+            )
+        return session.timestamp
+
+    @with_session
+    def create_batch(self, session, user, collection):
+        """Creates a batch in batch_uploads table"""
+        userid = user_key(user)
+        collectionid = self._get_collection_id(collection)
+        # TODO: raise ConflictError if the insert fails for an existing
+        # TODO: add the batchid skew (user % 10) or use the full microsecond TS
```
rfk: Lol it'd probably be quite nice if you didn't have to repeat that hackery for this batch implementation 😆

syncstorage/storage/spanner.py
```diff
@@ -0,0 +1,988 @@
...
+                        "collections",
+                        columns=["collectionid", "name"],
+                        values=[[collectionid, collection]]
+                    )
+                    return collectionid
+
+                # Run the transaction and get the collectionid we created
+                collectionid = self._database.run_in_transaction(
+                    create_collection)
+            except Exception as e:
```
rfk: Pretty broad `except` clause here; will it always be a particular type of exception e.g. `Aborted` like you use above for the locking stuff?

syncstorage/storage/sql/queries_spanner.py
```diff
@@ -0,0 +1,121 @@
...
+    sort = params.get("sort", None)
+    if sort == 'index':
+        order_args = [bso.c.sortindex.desc(), bso.c.id.desc()]
+    elif sort == 'oldest':
+        order_args = [bso.c.modified.asc(), bso.c.id.asc()]
+    else:
+        order_args = [bso.c.modified.desc(), bso.c.id.asc()]
+    # ...but unfortunately, sorting by "id" causes a significant slowdown on
+    # older versions of MySQL, so it's disabled by default and must be
+    # explicitly opted in to via config.
```
rfk: You can probably drop this mysql-specific workaround for spanner.

syncstorage/tests/functional/support.py
```diff
@@ -50,7 +51,14 @@ def _authenticate(self):
         self.user_id = random.randint(1, 100000)
         auth_policy = self.config.registry.getUtility(IAuthenticationPolicy)
         req = Request.blank(self.host_url)
-        creds = auth_policy.encode_hawk_id(req, self.user_id)
+        creds = auth_policy.encode_hawk_id(
+            req, self.user_id, extra={
+                # Include a hashed_fxa_uid to trigger uid/kid extraction
+                "hashed_fxa_uid": str(uuid.uuid4()),
```
rfk: I just noticed this annoyance as well while working on the loadtest update, I wonder if there's any better way to detect that "fxa_uid" is the new-style thing rather than the old hashed thing.  Maybe we could switch on the presence of an `fxa_kid` key instead, over in https://github.com/mozilla-services/server-syncstorage/blob/master/syncstorage/views/authentication.py#L182 ?

bbangert: So, all the validation logic in authentication.py is a bit of a mess. If you're ok with introducing a new dependency, I'd be happy to allow optional extraction/validation of more of these attributes without repeating the validation code over and over with Marshmallow (https://github.com/marshmallow-code/marshmallow).

syncstorage/storage/spanner.py
bbangert: Going to throw an error, as the database should be setup externally.

syncstorage/storage/spanner.py
bbangert: Yea, in Spanner its effectively a read-lock, in that all the reads have a single period of consistency.

syncstorage/storage/spanner.py
bbangert: Heh, copypasta gone awry.

syncstorage/storage/spanner.py
bbangert: It does, my reading of the code, specifically with how the lock is done on a per collection basis and access to this variable is only done when working under a single collection lock. This is slightly different to how the cached collection timestamps are handled in non-Spanner, as that use-case benefits from caching all of them since the Python process is the sole accessor of the database and can be aware that the actual values aren't changing out from under it.

I think if we wanted to assert this, we'd cache the collection in question with this timestamp, then verify on access that the collection being access is the same that had its timestamp cached. Note this is a session specific caching of the value though, not on the persistent storage object itself like the cached collection timestamps for non-Spanner.

syncstorage/storage/sql/queries_spanner.py
bbangert: Indeed.

syncstorage/tests/functional/support.py
rfk: Sounds good to me!

bbangert: I'm going to bring the marshmallow validation in with a separate PR/commit as it'll likely shuffle a lot of stuff around unrelated to Spanner.

syncstorage/storage/spanner.py
```diff
@@ -0,0 +1,987 @@
...
+    COLLECTION_CURRENT_TIMESTAMP,
+    COLLECTIONS_SIZES,
+    COLLECTIONS_COUNTS,
+    FIND_ITEMS,
+    STORAGE_SIZE,
+)
+
+logger = logging.getLogger(__name__)
+
+INT64_MAX = 63 ** 1 - 1
```
rfk: Something about this doesn't seem right...

```
>>> INT64_MAX = 63 ** 1 - 1
>>> INT64_MAX
62
```

syncstorage/storage/spanner.py
```diff
@@ -0,0 +1,987 @@
...
+    """
+    return ":".join([user["fxa_uid"], user["fxa_kid"]])
+
+
+def getq(query):
+    kwargs = {}
+    if "%(bso)s" in query:
+        kwargs["bso"] = "bso"
+    if "%(bui)s" in query:
+        raise RuntimeError("XXX: batches")
```
rfk: nit: please raise a more descriptive error message here; I assume it should never happen in practice because we don't use those queries, but will be good to have more context if it ever does.

syncstorage/storage/spanner.py
```diff
@@ -0,0 +1,987 @@
...
+    if "%(ids)s" in query:
+        kwargs["ids"] = "@ids"
+    if kwargs:
+        query = query % kwargs
+    return query.replace(":", "@")
```
rfk: A part of me wonders if there are any edge-cases where this will replace something other than a named parameter; could we use a regex replacement to be more confident about what this matches?

syncstorage/storage/spanner.py
```diff
@@ -0,0 +1,987 @@
...
+                self._collections_by_id[id] = name
+
+        # A thread-local to track active sessions.
+        self._tldata = threading.local()
+
+    def _lock_write_precondition(self, session, user, collection):
+        """'Lock' the precondition on the collection
+
+        Get the timestamp of the collection in use, utilizing it as
+        precondition for
```
rfk: nit: extra newline in the docstring here?

syncstorage/storage/spanner.py
```diff
@@ -0,0 +1,987 @@
...
+            param_types={"userid": param_types.STRING,
+                         "collectionid": param_types.INT64}
+        ).one_or_none()
+        if result:
+            current_ts, last_modified = result
+            session.collection_ts = dt2ts(last_modified)
+            session.timestamp_precise = current_ts
+            session.timestamp = dt2ts(current_ts)
+            # Forbid the write if it would not properly incr the timestamp.
+            if session.collection_ts >= dt2ts(current_ts):
```
rfk: nit: you could just use `session.timestamp` here rather than calling `dt2ts` twice.

syncstorage/storage/sql/queries_spanner.py
```diff
@@ -0,0 +1,114 @@
...
+    name            STRING(MAX) NOT NULL,
+) PRIMARY KEY (collectionid);
+
+CREATE TABLE batches (
+    userid          STRING(MAX) NOT NULL,
+    collection      INT64 NOT NULL,
+    id              TIMESTAMP NOT NULL,
+    bsos            STRING(MAX) NOT NULL,
+    expiry          TIMESTAMP NOT NULL
+) PRIMARY KEY (userid, collection, id);
```
rfk: IIUC, this needs an `ON DELETE CASCADE` in order for this earlier comment to be true:

```
# Deleting user_collections will wipe out all BSO's and batches
# for the user due to cascading delete
```

syncstorage/storage/spanner.py
```diff
@@ -0,0 +1,987 @@
...
+
+    def _find_items(self, session, user, collection, **params):
+        """Find items matching the given search parameters."""
+        userid = user_key(user)
+        collectionid = self._get_collection_id(collection)
+        bind = {"userid": userid,
+                "collectionid": collectionid}
+        bind_types = {"userid": param_types.STRING,
+                      "collectionid": param_types.INT64}
+        ts = get_timestamp()
```
rfk: If available, should this use a timestamp from the `session` object for consistency?

syncstorage/storage/spanner.py
```diff
@@ -0,0 +1,987 @@
...
+
+        # Convert timestamp types
+        for ts_var in ["newer", "newer_eq", "older", "older_eq", "ttl"]:
+            if ts_var in params:
+                bind[ts_var] = ts2dt(params[ts_var])
+                bind_types[ts_var] = param_types.TIMESTAMP
+
+        # Generate the query
+        query = FIND_ITEMS(bso, params)
+        query = str(query.compile()).replace(":", "@")
```
rfk: As above, is there any chance of this `.replace` overzealously matching something it shouldn't? I can't think of any concrete example off the top of my head, just wondering out loud...

syncstorage/storage/spanner.py
```diff
@@ -0,0 +1,987 @@
...
+                userid,
+                collectionid,
+                item["id"],
+                item
+            )
+            session.transaction.update(
+                "bso",
+                columns=cols,
+                values=[vals]
+            )
```
rfk: Oh, interesting, we have to issue a separate update query for each item that is changed?  Is there any way to build all the update statements and issue them together in a single command, rather than dispatching each as an individual query to the db?

(Not a blocker, just something to think about for a future iteration)

syncstorage/storage/spanner.py
```diff
@@ -0,0 +1,987 @@
...
+                              items):
+        """Inserts items into batch_upload_items"""
+        userid = user_key(user)
+        collectionid = self._get_collection_id(collection)
+        batchid = ts2dt(batchid / 1000.0)
+        items_json = [json.dumps(item) for item in items]
+        bsos = "%s%s" % ("\n".join(items_json), "\n" if items_json else "")
+        result = session.transaction.execute_update(
+            """\
+            UPDATE batches SET bsos = CONCAT(bsos, @bsos)
```
rfk: Can we detect the case where `bsos` exceeds the max size limit on a string, and error out cleanly?

syncstorage/storage/sql/queries_spanner.py
```diff
@@ -0,0 +1,114 @@
+from sqlalchemy.sql import select, bindparam
```
rfk: It's a little weird to have this one file of spanner-specific stuff over here under `storage/sql`, and then have the rest of the spanner backend in a different directory?  Is it worth making a `storage/spanner/` directory that can host them both?

syncstorage/tests/functional/support.py
```diff
@@ -219,10 +227,12 @@ def _authenticate(self):
 
     # Now use the unittest2 runner to execute them.
     suite = unittest2.TestSuite()
-    suite.addTest(unittest2.makeSuite(LiveTestCases))
+    test_prefix = os.environ.get("SYNC_TEST_PREFIX", "test")
+    suite.addTest(unittest2.makeSuite(LiveTestCases, prefix=test_prefix))
```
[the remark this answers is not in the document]
pjenvey: this was a quick way to have the runner run tests individually e.g. `SYNC_TEST_PREFIX=test_get_collection`

syncstorage/storage/spanner.py
bbangert: There's no way for a user to insert data into our query, as we're using bind params for all variables, and the existing queries use : purely for binding (or they'd fail as well). So the only way I could see that this would cause problems is if the original queries have invalid variable bindings.

syncstorage/storage/sql/queries_spanner.py
pjenvey: we're just going to remove this comment, it's outdated. AFAICT the original impl. doesn't actually delete batches when deleting storage so we match its behavior.

we originally interleaved these tables which would have caused the cascade but removed it -- i've added this comment to the DDL:
```
-- batches is not interleaved w/ user_collections because batches can
-- be created before there's a row in that table (test_storage has a
-- test that triggers this). spanner requires the parent's row to
-- be present before the child's
```

syncstorage/storage/spanner.py
bbangert: We can set a pessimistically low limit on batches to prevent it, that's about it I think.

syncstorage/storage/spanner.py
bbangert: Filed https://github.com/mozilla-services/server-syncstorage/issues/99.

syncstorage/storage/spanner.py
bbangert: Adding a max_total_bytes setting of 2.2 MB to ensure there's JSON serialization space around it and it will still fit in 2.5 MB.

PR #102, opened 2019-02-14
Makes gunicorn --worker-class configurable.

Micheletto:
Needed for bug [1520091](https://bugzilla.mozilla.org/show_bug.cgi?id=1520091).

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -15,7 +15,7 @@ case "$1" in
         exec gunicorn \
             --paste "$_SETTINGS_FILE" \
             --bind ${HOST-127.0.0.1}:${PORT-8000}\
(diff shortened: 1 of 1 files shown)


docker-entrypoint.sh
```diff
@@ -15,7 +15,7 @@ case "$1" in
         exec gunicorn \
             --paste "$_SETTINGS_FILE" \
             --bind ${HOST-127.0.0.1}:${PORT-8000}\
-            --worker-class mozsvc.gunicorn_worker.MozSvcGeventWorker \
+            --worker-class ${WORKER_CLASS} \
```
bbangert: Should a default be set?

Micheletto: > Should a default be set?

Sure.

PR #103, opened 2019-02-18
chore: Add /__lbheartbeat__ route

### LINKED ISSUE #101: Add a __lbheartbeat__ endpoint.
Now that Sync is going to be fronted by a load balancer we will need an \_\_lbheartbeat\_\_ endpoint as described [here](https://mana.mozilla.org/wiki/display/SVCOPS/New+Service+Guide).  The purpose is to avoid marking nodes or instances bad because an upstream service (storage, etc.) is down.

rfk:
Fixes #101; @pjenvey r?

Code lifted directly from the exiting route in tokenserver.

### DIFF
--- a/syncstorage/views/__init__.py
+++ b/syncstorage/views/__init__.py
@@ -134,6 +134,25 @@ def get_site_root(request):
     return "It Works!  SyncStorage is successfully running on this host."
 
 
+# The /__lbheartbeat__ route is used by mozsvc load-balancer infra
+# to detect when a webhead is out of service.
+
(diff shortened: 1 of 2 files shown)


syncstorage/views/__init__.py
```diff
@@ -134,6 +134,25 @@ def get_site_root(request):
     return "It Works!  SyncStorage is successfully running on this host."
 
 
+# The /__lbheartbeat__ route is used by mozsvc load-balancer infra
+# to detect when a webhead is out of service.o
```
vladikoff: what's a `service.o`?

rfk: lol, thanks :-)

PR #115, opened 2019-04-01
feat: force specify the higher bound for basic modified ranges

pjenvey:
this doesn't seem to affect FIND_ITEM's query plan but including it
just in case (spanner's best practices suggests including it)

### DIFF
--- a/syncstorage/storage/sql/queries_spanner.py
+++ b/syncstorage/storage/sql/queries_spanner.py
@@ -1,4 +1,4 @@
-from sqlalchemy.sql import select, bindparam
+from sqlalchemy.sql import select, bindparam, text
 
 DATABASE_CREATE_DDL = """\
 CREATE TABLE user_collections (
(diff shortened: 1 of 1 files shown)


syncstorage/storage/sql/queries_spanner.py
```diff
@@ -89,8 +89,10 @@ def FIND_ITEMS(bso, params):
         query = query.where(bso.c.id.in_(params.get("ids")))
     if "newer" in params:
         query = query.where(bso.c.modified > bindparam("newer"))
+        query = query.where(bso.c.modified <= text("CURRENT_TIMESTAMP()"))
```
rfk: I can imagine future readers wondering why this is here, could you please add a comment to explain the rationale?

bbangert: Or link to relevant Spanner docs?

PR #116, opened 2019-04-17
fix(deps): Update SQLAlchemy to latest version

rfk:
Resolves a security warning (which I don't believe affects us in practice, but seems worth updating to avoid spurious github alerts in any case).


rfk:
Actually this wanted to pull in a few other updated dependencies, so I just went ahead and updated them all and did a fresh freeze into the requirements file.

PR #119, opened 2019-05-15
feat: optionally split FIND_ITEMS' query in 2

### LINKED ISSUE #118: Add a flag for splitting FIND_ITEMS into multiple queries
Under stess testing the LastModified secondary index produces slower FIND_ITEMS results w/ higher Spanner CPU usage than not using it at all, specifically on larger result sets generated by the load tester.

FIND_ITEMS implicitly joins the index against the BSO table (to load columns not included in the index). Spanner's query planner produces a _Cross Apply_ w/ an _Input_ of the results of the index query and a _Map_ on the bso table. Each _Map_ produces an individual "execution" and each execution completes rather quickly, but on larger result sets begins to add up.

Screen Shot 2019-04-04 at 2 30 04 PM

We can avoid all the individual executions by doing the index query first and passing its results to a separate query to load the other columns.

Let's add a flag to test this under load. 

We should expect this secondary index to work more efficiently on typical FIND_ITEMS work loads producing a smaller number of results. The goal here is to utilize it for this sake while still producing reasonable response times/cpu usage with larger results as well.

pjenvey:
via _bsolm_index_separate = true

Closes #118

### DIFF
--- a/syncstorage/storage/spanner.py
+++ b/syncstorage/storage/spanner.py
@@ -53,6 +53,9 @@ INT64_MAX = 2 ** 63 - 1
 
 EPOCH = datetime.datetime.utcfromtimestamp(0)
 
+# a bogus FIND_ITEMS bind param
+BOGUS_ID = object()
+
 
 def dt2ts(dt):
     """Convert a Python datetime to seconds"""
@@ -121,8 +124,18 @@ class SpannerStorage(SyncStorage):
             target_size=int(dbkwds.get("pool_size", 100))
         )
         self._database = self._instance.database(database_id, pool=pool)
+
         # for debugging: support disabling FORCE_INDEX=BsoLastModified
         self._force_bsolm_index = dbkwds.get("_force_bsolm_index", True)
+        # split FIND_ITEMS' query in 2 (first query BsoLastModified
+        # then query bso with its results)
+        self._bsolm_index_separate = dbkwds.get(
+            "_bsolm_index_separate",
+            False
+        )
+        if self._bsolm_index_separate:
+            assert self._force_bsolm_index, \
+                "_bsolm_index_separate requires _force_bsolm_index"
 
         self.standard_collections = standard_collections
         if self.standard_collections and dbkwds.get("create_tables", False):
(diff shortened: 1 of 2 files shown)


syncstorage/storage/spanner.py
```diff
@@ -358,13 +371,68 @@ def _find_items(self, session, user, collection, **params):
                 bind_types[ts_var] = param_types.TIMESTAMP
 
         # Generate the query
+        fields = params.get("fields")
+        offset_params = params
+
+        if self._bsolm_index_separate and fields is None:
+            # Split the get_items query (TODO: get_item_ids could
+            # reference solely the index) query in 2:
+            # #1: query BsoLastModified directly for bso ids
```
bbangert: What's the intended goal of using this 2 query version? I assume its an optimization to avoid the multiple executions that occur otherwise?

pjenvey: Yea, see my description in the issue #118

pjenvey: and if this solution looks alright under load testing I'd like to relay those results back to Google, in case they have an alternative or a better explanation for the existing single query's perf

syncstorage/storage/spanner.py
```diff
@@ -53,6 +53,9 @@
 
 EPOCH = datetime.datetime.utcfromtimestamp(0)
 
+# a bogus FIND_ITEMS bind param
+BOGUS_ID = object()
```
jrconlin: Is this used (or planned to be used) anywhere outside of the SpannerStorage.__init__()?

pjenvey: it's only used in _find_items as basically a placeholder

syncstorage/storage/spanner.py
```diff
@@ -358,13 +371,68 @@ def _find_items(self, session, user, collection, **params):
...
+                del bind["limit"]
+                del bind_types["limit"]
+            if offset is not None:
+                # restore this later
+                del params["offset"]
+                del bind["offset"]
+                del bind_types["offset"]
+            # simiarly modified ranges/ttl aren't needed in #2
+            for param in ("newer", "newer_eq", "older", "older_eq", "ttl"):
+                params.pop(param, None)
```
rfk: There's quite a big chunk of code in this `if` block, I wonder if it could be factored out into a helper method of some sort.

syncstorage/storage/spanner.py
```diff
@@ -358,13 +371,68 @@ def _find_items(self, session, user, collection, **params):
                 bind_types[ts_var] = param_types.TIMESTAMP
 
         # Generate the query
+        fields = params.get("fields")
+        offset_params = params
+
+        if self._bsolm_index_separate and fields is None:
```
rfk: It's not obvious why `and fields is None` here, is it to ensure we only apply this logic in `get_items` rather than `get_item_ids`?  If so, I wonder if it's worth refactoring `_find_items` into a couple of helper methods and pushing that logic up into `get_items`, like:

```
def get_items(...):
    if self._bsolm_index_separate:
        (params, ids) = self._do_bsolm_separate_query_to_get_the_ids()
    return self._find_items(**params)

def get_item_ids(...):
    return self._find_items(...)
```

Depends how much effort you want to invest in polishing this though, as opposed to just getting it out in stage to loadtest.

syncstorage/storage/spanner.py
```diff
@@ -358,13 +371,68 @@ def _find_items(self, session, user, collection, **params):
...
+                param_types=bind_types,
+            )
+            ids = [row[0] for row in list(result)]
+
+            # keep the orig. for encode_next_offset
+            offset_params = params.copy()
+
+            # Setup a 'id IN (:id_1)' bind param for #2 (replacing it
+            # below)
+            params["ids"] = [BOGUS_ID]
```
rfk: What stops us from just putting the returned `ids` in `params["id"]`, and basically just pretending that the caller had requested those ids specifically?

rfk:
@pjenvey do we need to get this merged before loadtesting, or can we do an initial test from this branch before deciding whether to push ahead with it or not?

syncstorage/storage/spanner.py
pjenvey: That's right, to only apply it to get_items. I like the refactor suggestion but it's a little further complicated by encode_next_offset's need for the original params (`offset_params`).

I will keep it in mind if this is revisited. This is definitely a "get it out to stage and see" patch

syncstorage/storage/spanner.py
pjenvey: This is specifically to utilize spanner's `UNNEST` (what it's replaced with below) for query 2.

I do have a TODO for the FIND_ITEMS/sqlalchemy generation piece to do this for us. params["ids"] from the client and/or this case should *both* utilize UNNEST. Unfortunately coercing the sqlalchemy level to generate it for us is a bit involved, thus the current str replace.

Another quick solution for this is having FIND_ITEMS handle UNNEST by returning a str with the appropriate replacements (also not ideal, further straying from the original FIND_ITEMS code..) 🤷‍♂

pjenvey:
Sorry I missed the last comment until just now -- I can certainly tag this from the branch for an initial load test if more is needed to merge this PR

rfk:
> I can certainly tag this from the branch for an initial load test if more is needed to merge this PR

Whatever seems simplest for you. I'm happy to r+ with followup bugs for refactors, but if we're not sure yet whether we'll even keep this code, deploying from a branch feels like it might be the simpler path forward overall.

PR #122, opened 2019-06-24
feat: Add delayed retry on 500/503 responses.

### LINKED ISSUE #121: Add delayed retry on 500/503 responses
Several tests fail due to SQL lock contention on the server. This is because lock is derived from the server request millisecond, which may be processed too quickly.

The UA currently retries after a delay, the tests should replicate this behavior.

jrconlin:
Closes #121

### DIFF
--- a/syncstorage/tests/functional/test_storage.py
+++ b/syncstorage/tests/functional/test_storage.py
@@ -86,9 +86,7 @@ class TestStorage(StorageFunctionalTestCase):
         try:
             return func(*args, **kwargs)
         except webtest.AppError as ex:
-            # TODO: Temporarily accept 500 & 503 due to bug in syncstorage-rs
-            # replace 500 with 409 in the future.
-            if "500 " not in ex.message and "503 " not in ex.message:
+            if "409 " not in ex.message and "503 " not in ex.message:
                 raise ex
             time.sleep(0.01)
             return func(*args, **kwargs)
@@ -1576,7 +1574,6 @@ class TestStorage(StorageFunctionalTestCase):
         self.assertTrue('max_request_bytes' in limits)
 
         endpoint = self.root + '/storage/xxx_col2?batch=true'
-
         # There are certain obvious constraints on these limits,
         # violations of which would be very confusing for clients.
 
@@ -1635,7 +1632,6 @@ class TestStorage(StorageFunctionalTestCase):
             'X-Weave-Bytes': str(limits['max_post_bytes'] + 1)
         }, status=400)
         self.assertEquals(res.json, WEAVE_SIZE_LIMIT_EXCEEDED)
-
         bsos = [
             {'id': 'little', 'payload': 'XXX'},
             {'id': 'big', 'payload': 'X' * (limits['max_post_bytes'] - 3)}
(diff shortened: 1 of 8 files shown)


jrconlin:
Since `get` doesn't impose the lock requirement, it's not really required to go through the retry method wrapper.

syncstorage/tests/functional/test_storage.py
```diff
@@ -73,27 +73,37 @@ def _switch_user(self):
         finally:
             self.root = orig_root
 
+    def retry(self, method, *args, **kwargs):
```
pjenvey: Another option is monkey patching self.app.do_request (which is already done to setup auth tokens here: https://github.com/mozilla-services/server-syncstorage/blob/master/syncstorage/tests/functional/support.py#L42). Could only apply if `req.method not in ('GET', 'HEAD')`. It would reduce this patch to only a few lines

pjenvey: I think the loop should strictly match what [the decorator does](https://github.com/mozilla-services/server-syncstorage/blob/master/syncstorage/views/decorators.py#L62): just 1 retry.

jrconlin: I'm not a fan of monkey patching, mostly because it's too easy to lose the patch unless you know to look for it. 

I'll fix up `retry` to use the same time interval and single retry like you pointed out.

syncstorage/tests/functional/test_storage.py
```diff
@@ -73,27 +73,37 @@ def _switch_user(self):
...
+        func = getattr(self.app, method)
+        try:
+            return func(*args, **kwargs)
+        except webtest.AppError as ex:
+            if "500 " not in ex.message and "503 " not in ex.message:
```
pjenvey: Oh and one more: we should only retry on 503s. The current issue w/ syncstorage-rs triggering 500s in some conflicting lock_for_write cases is a bug

syncstorage/tests/functional/test_storage.py
rfk: I'm personally not a huge fan of string-named method calls like this, but it's purely a stylistic thing so I won't push back on it.  You could also consider making wrappers like `self.delete_with_retry`, self.post_json_with_retry` or similar that might read a little more cleanly.

jrconlin: I'll be honest, the `retry(func,..` was mostly just to make the find/replace a lot easier, and to limit the patch to just one bottleneck function. I can see the value of clearly declared function names, though.

syncstorage/tests/functional/test_storage.py
jrconlin: Ok. I'm not a super fan of scraping out the error code from the message like this. I may try something cleaner. 

I'll agree that the `500` is there mostly because the rust is currently returning the wrong error. I'm hoping to have that fixed up separately. I may just keep this branch a WIP until it is.

pjenvey: Unfortunately webtest.AppError somewhat swallows responses, the str match is the only way to check the error code on it

With enough effort you could probably pass along `expect_errors=True`, getting back a response object instead, then [manually check for bad status codes](https://github.com/Pylons/webtest/blob/master/webtest/app.py#L663), reraising the AppError yourself. Not sure it's worth bothering

PR #133, opened 2019-12-30
Add "extension-storage" to list of standard collections.

### LINKED ISSUE #128: New extension-storage collection will need to be in STANDARD_COLLECTIONS
(And anywhere else along the same lines). We're working on a sync15-based engine for extension-storage, moving it from Kinto. Probably some (small) number of things might be be desirable to change in this repo to prepare for that, although the only one I know of is stuff like the STANDARD_COLLECTIONS list.

rfk:
Fixes #128; r? @thomcc to confirm the correct name for the collection.

### DIFF
--- a/syncstorage/storage/sql/__init__.py
+++ b/syncstorage/storage/sql/__init__.py
@@ -44,11 +44,13 @@ logger = logging.getLogger(__name__)
 # For efficiency, it's possible to use fixed pre-determined IDs for
 # common collection names.  This is the canonical list of such names.
 # Non-standard collections will be allocated IDs starting from the
-# highest ID in this collection.
+# `FIRST_CUSTOM_COLLECTION_ID`, which must be larger than the highest
+# key in this collection.
 STANDARD_COLLECTIONS = {1: "clients", 2: "crypto", 3: "forms", 4: "history",
                         5: "keys", 6: "meta", 7: "bookmarks", 8: "prefs",
                         9: "tabs", 10: "passwords", 11: "addons",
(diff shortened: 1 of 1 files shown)


pjenvey:
The last time we added to the std collections @Micheletto noted the ["extension-storage" collection already existed](https://bugzilla.mozilla.org/show_bug.cgi?id=1364005#c3) with what I assume a collection_id > 100 (not 14).

That really complicates adding it to the std list. Should we even bother?

rfk:
> "extension-storage" collection already existed 

Oh wow. I don't know what this might be used for, but it's definitely something other than what @thomcc is about to start using it for. Quite aside from the collection-id issue, whatever is using that collection name seems likely to have problems when the new extension-storage stuff starts clobbering its data in that collection. And we may start seeing impenetrable bug reports when it starts clobbering data that we wrote into that collection.

ISTM we should either convince ourselves that whatever was using that collection name is no longer a significant concern (in which case we can take it over) or we should pick a different name for our own collection. @thomcc thoughts?

rfk:
Per meeting: WONTFIX; @tublitzed is going to follow up on whether we can use the existing "extension-storage" name.

tublitzed:
Per email with @Micheletto and @thomcc we opted to just use a new name. We're going to  [investigate existing collections names](https://bugzilla.mozilla.org/show_bug.cgi?id=1617676) however just to understand how much we should worry about this moving forward/if we need a naming strategy/or to purge collections/etc.

PR #136, opened 2020-01-23
feat: Add table to cross check user migration status

jrconlin:
This branch adds support to look in a `migration` table for a user's
`fxa_uid` record. If it is found, the server will return a 5xx error to
the client preventing sync operations.

NOTE: IT IS IMPORTANT THAT THE SERVER NOT RETURN A 2xx RESULT TO ANY
CLIENT IN MIGRATION STATUS UNLESS THE `meta`/`global` BSO IS RESET.

The `migration` table is created by the `user_migration` script, which
is part of `syncstorage-rs`.

### DIFF
--- a/syncstorage/storage/sql/__init__.py
+++ b/syncstorage/storage/sql/__init__.py
@@ -132,6 +132,7 @@ class SQLStorage(SyncStorage):
     """
 
     def __init__(self, sqluri, standard_collections=False, **dbkwds):
+        self.allow_migration = dbkwds.get("allow_migration", False)
         self.sqluri = sqluri
         self.dbconnector = DBConnector(sqluri, **dbkwds)
         self._optimize_table_before_purge = \
@@ -408,12 +409,13 @@ class SQLStorage(SyncStorage):
         guessing if a migration has not yet started.
 
         """
-        userid = user.get("fxa_uid")
-        if userid is not None:
-            res = session.query_fetchone("MIGRATION_CHECK", {
-                "fxa_uid": userid
-            })
-            return res is not None
+        if self.allow_migration:
+            userid = user.get("fxa_uid")
+            if userid is not None:
+                res = session.query_fetchone("MIGRATION_CHECK", {
+                    "fxa_uid": userid
+                })
+                return res is not None
         return False
 
     def _row_to_bso(self, row, timestamp):
--- a/syncstorage/storage/sql/__init__.py
+++ b/syncstorage/storage/sql/__init__.py
@@ -397,6 +397,25 @@ class SQLStorage(SyncStorage):
             "next_offset": next_offset,
         }
 
+    @with_session
+    def is_migrating(self, session, user):
+        """Returns True if the user is migrating.
+
+        This should probably NOT be a cached call, since migration
+        may start at any time during a client sync.
+
+        any error will percolate up as a 500, which will be safer than
+        guessing if a migration has not yet started.
+
+        """
+        userid = user.get("fxa_uid")
+        if userid is not None:
+            res = session.query_fetchone("MIGRATION_CHECK", {
+                "fxa_uid": userid
+            })
+            return res is not None
+        return False
+
     def _row_to_bso(self, row, timestamp):
         """Convert a database table row into a BSO object."""
         item = dict(row)
--- a/syncstorage/storage/sql/dbconnect.py
+++ b/syncstorage/storage/sql/dbconnect.py
@@ -379,7 +379,10 @@ class DBConnector(object):
             collections.create(self.engine, checkfirst=True)
             user_collections.create(self.engine, checkfirst=True)
             batch_uploads.create(self.engine, checkfirst=True)
-            migration.create(self.engine, checkfirst=True)
+            # only create migration if "[storage]:allow_migration" flag
+            # is set
+            if kwds.get("allow_migration", False):
+                migration.create(self.engine, checkfirst=True)
             if not self.shard:
                 bso.create(self.engine, checkfirst=True)
                 bui.create(self.engine, checkfirst=True)
--- a/syncstorage/storage/sql/dbconnect.py
+++ b/syncstorage/storage/sql/dbconnect.py
@@ -102,6 +102,22 @@ user_collections = Table(
 )
 
 
+# Table listing migration state for users
+#
+# This table contains the users in progress for migration. This table may
+# be updated by the external user migration script
+# A fxa_uid appearing in this table should cause the server to return a 5xx
+# response. It's important that the server NEVER return a 2xx response.
+
+migration = Table(
+    "migration",
+    metadata,
+    Column("fxa_uid", String(255), primary_key=True, nullable=False),
+    Column("started_at", BigInteger, nullable=False),
+    Column("state", String(32))  # unknown/NULL, in_progress, complete
+)
+
+
 # Column definitions for BSO storage table/tables.
 #
 # This list class defines the columns used for storage of BSO records.
@@ -363,6 +379,7 @@ class DBConnector(object):
             collections.create(self.engine, checkfirst=True)
             user_collections.create(self.engine, checkfirst=True)
             batch_uploads.create(self.engine, checkfirst=True)
+            migration.create(self.engine, checkfirst=True)
             if not self.shard:
                 bso.create(self.engine, checkfirst=True)
                 bui.create(self.engine, checkfirst=True)
--- a/syncstorage/tests/functional/support.py
+++ b/syncstorage/tests/functional/support.py
@@ -45,7 +45,7 @@ class StorageFunctionalTestCase(FunctionalTestCase, StorageTestCase):
         orig_do_request = self.app.do_request
         self.app.do_request = new_do_request
 
-    def _authenticate(self):
+    def _authenticate(self, fxa_uid=None):
         # For basic testing, use a random uid and sign our own tokens.
         # Subclasses might like to override this and use a live tokenserver.
         self.user_id = random.randint(1, 100000)
@@ -55,7 +55,7 @@ class StorageFunctionalTestCase(FunctionalTestCase, StorageTestCase):
             req, self.user_id, extra={
                 # Include a hashed_fxa_uid to trigger uid/kid extraction
                 "hashed_fxa_uid": str(uuid.uuid4()),
-                "fxa_uid": str(uuid.uuid4()),
+                "fxa_uid": fxa_uid or str(uuid.uuid4()),
                 "fxa_kid": str(uuid.uuid4()),
             }
         )
--- a/syncstorage/tests/test_sql.py
+++ b/syncstorage/tests/test_sql.py
@@ -31,7 +31,7 @@ class TestSQLStorage(StorageTestCase, StorageTestsMixin):
 
     def setUp(self):
         super(TestSQLStorage, self).setUp()
-        settings = self.config.registry.settings
+        settings = self.config.registry.setting
         self.storage = load_storage_from_settings("storage", settings)
 
     def test_no_create(self):
@@ -216,6 +216,10 @@ class TestSQLStorage(StorageTestCase, StorageTestsMixin):
             )
 
     def test_migrating(self):
+        settings = self.config.registry.setting
+        settings['allow_migration'] = True
+        self.storage = load_storage_from_settings("storage", settings)
+
         migrating_id = str(uuid.uuid4())
         self.assertFalse(
             self.storage.is_migrating(
@@ -225,6 +229,15 @@ class TestSQLStorage(StorageTestCase, StorageTestsMixin):
         )
         self._set_migrating_state(migrating_id, 'in_progress')
         self.assertTrue(
+            self.storage.is_migrating(
+                {"fxa_uid": migrating_id,
+                 "uid": 1}
+            )
+        )
+
+        self.storage.allow_migration = False
+        self._set_migrating_state(migrating_id, 'in_progress')
+        self.assertFalse(
             self.storage.is_migrating(
                 {"fxa_uid": migrating_id,
                  "uid": 1}
--- a/syncstorage/tests/test_sql.py
+++ b/syncstorage/tests/test_sql.py
@@ -4,6 +4,7 @@
 
 import time
 import threading
+import uuid
 
 from mozsvc.plugin import load_and_register
 from mozsvc.tests.support import get_test_configurator
@@ -197,3 +198,35 @@ class TestSQLStorage(StorageTestCase, StorageTestsMixin):
         self.assertEquals(count_items(), 5)
         self.assertEquals(len(self.storage.get_items(_USER, "col")["items"]),
                           5)
+
+    def _set_migrating_state(self, id, state):
+        with self.storage.dbconnector.connect() as connect:
+            connect.execute(
+                """
+                INSERT INTO migration
+                    (fxa_uid, started_at, state)
+                VALUES
+                    (:id, :time, :state)
+                /* queryName=migrate */""",
+                params={
+                    "id": id,
+                    "time": int(time.time()),
+                    "state": state
+                }
+            )
+
+    def test_migrating(self):
+        migrating_id = str(uuid.uuid4())
+        self.assertFalse(
+            self.storage.is_migrating(
+                {"fxa_uid": migrating_id,
+                 "uid": 1}
+            )
+        )
+        self._set_migrating_state(migrating_id, 'in_progress')
+        self.assertTrue(
+            self.storage.is_migrating(
+                {"fxa_uid": migrating_id,
+                 "uid": 1}
+            )
+        )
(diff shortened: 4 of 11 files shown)


syncstorage/storage/sql/dbconnect.py
```diff
@@ -102,6 +102,21 @@ def __new__(cls, *args, **kwds):
...
+# This table contains the users in progress for migration. This table may
+# be updated by the external user migration script
+# A UID appearing in this table should cause the server to return a 5xx
+# response. It's important that the server NEVER return a 2xx response.
+
+migration = Table(
+    "migration",
+    metadata,
+    Column("fxa_uid", String(255), primary_key=True, nullable=False),
+    Column("started", BigInteger, nullable=False)
```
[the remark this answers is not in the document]
rfk: nit: it took me a long time to work out that you intended `started` to contain a timestamp, please consider calling this `started_at` or `start_time` or similar.

syncstorage/tests/test_sql.py
```diff
@@ -197,3 +198,21 @@ def count_items():
...
+
+    def test_migrating(self):
+        migrating_id = str(uuid.uuid4())
+        with self.storage.dbconnector.connect() as connect:
+            sql = """
+                INSERT INTO migration
+                    (fxa_uid, started)
+                VALUES
+                    ('{}', {})
+                /* queryName=migrate */""".format(
```
rfk: nit: I know it's just a test, but you could use sql parameters here rather than string formatting:

```
connect.execute(sql, params={"fxa_uid": migrating_id, "started": int(time.time())})
```

syncstorage/views/decorators.py
```diff
@@ -192,3 +196,19 @@ def with_collection_lock(viewfunc, request):
...
+    """Check if a given user is in a migration state, return 503.
+
+    NOTE: once a user migration has started DO NOT RETURN A 2xx
+    WITHOUT REMOVING THE meta/global RECORD!
+
+    """
+    storage = request.validated["storage"]
+    if storage.is_migrating(request.user):
+        raise HTTPServiceUnavailable(
+            headers={"Retry-After": MIGRATION_RETRY_AFTER}
```
rfk: Elsewhere in this codebase we use `str` to ensure that header values are strings before returning them, and I have vague recollections of errors that happen if you try to return an int in a header like this. According to #137 this works fine passing it as an integer, but you might like to use `str(MIGRATION_RETRY_AFTER)` here for consistency with other retry-header-setting code.

rfk:
Please also consider the additional test from https://github.com/mozilla-services/server-syncstorage/pull/137

syncstorage/views/decorators.py
```diff
@@ -28,6 +28,10 @@
 # How long the client should wait before retrying a conflicting write.
 RETRY_AFTER = 10
 
+# How long a client should wait to retry. It won't succeed, but should
+# return a 4xx error punting them to the new server.
```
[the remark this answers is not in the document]
jrconlin: As I Understand, when a client gets a 4xx error, it needs to get a new token because it's been "reassigned". The migration script reaches into the tokenserver db as well to "assign" the user to spanner. I was originally thinking that I'd just wait until the client's token expires and the client naturally goes and checks for the new assignment, since that would potentially give the longest migration period.  That said, going to add a column to `migration` that is the `state`.

syncstorage/views/__init__.py
```diff
@@ -179,6 +180,7 @@ def get_lbheartbeat(request):
 
 @info_timestamps.get(accept="application/json", renderer="sync-json",
                      acl=expired_token_acl)
+@check_migration
```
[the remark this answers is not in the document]
jrconlin: This might be a styling nit for me. I'm not a huge fan of decorators to begin with because they secretly alter code flow and/or variables. Burying them inside a meta decorator that may or may not be applied doesn't really help, unless you keep what `default_decorators` does at top of mind at all times. 

That said, we'd need to do this check for any call the client might make which could return a 2xx result. I *could* add it to the `default_decorators` meta, and then add to the additional ones, but since this is kind of a semi-hack specific to the migration, I'd rather be very explicit about what's going on.

syncstorage/storage/sql/__init__.py
```diff
@@ -397,6 +397,24 @@ def _find_items(self, session, user, collection, **params):
...
 
+    @with_session
+    def is_migrating(self, session, user):
+        """Returns if the user is migrating.
+
+        This should probably NOT be a cached call, since migration
+        may start at any time during a client sync.
+
+        """
+        userid = user.get("fxa_uid", user["uid"])
```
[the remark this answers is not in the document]
jrconlin: I was mostly thinking about self hosters here, but you're right, they shoudn't be part of this at this time.

syncstorage/views/decorators.py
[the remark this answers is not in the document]
rfk: Gotcha. I think it would be OK to do as you suggest and just wait for the token to naturally expire, if you don't want the additional complexity (but in that case we should change the comment above to avoid confusion).

syncstorage/views/__init__.py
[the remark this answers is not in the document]
rfk: I'm fine with leaving it separate if that fits best with your mental model 👍

syncstorage/views/decorators.py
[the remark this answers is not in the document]
rfk: Although do be aware that we can under some circumstances accept expired tokens for up to 2 hours after they expire, for reasons:

https://github.com/mozilla-services/server-syncstorage/blob/a7649edc1f02e529beb85a24c329f1c5ce108f4d/syncstorage/views/authentication.py#L18

I'm not sure we ever actually measured whether that was worth it in practice, we could consider removing it if we want users to hurry up and discover their migrated status.

syncstorage/tests/test_sql.py
```diff
@@ -197,3 +198,21 @@ def count_items():
         self.assertEquals(count_items(), 5)
         self.assertEquals(len(self.storage.get_items(_USER, "col")["items"]),
                           5)
+
+    def test_migrating(self):
+        migrating_id = str(uuid.uuid4())
```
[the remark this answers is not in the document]
jrconlin: hrm, this is a bit tricky. The best way to test this is to add a test under `tests/functional/test_storage.py` which would add a record to a migration table and then test that the middleware layer returns a 503 when a matching fxa_uid record request is made. The problem is that there's a significant abstraction layer between the test architecture and the wsgi app such that I can't get to the wsgi storage from inside of the test framework. Since it's using `sqlite:////memory` that also complicates things. 

Another problem is that the `fxa_uid` is set to a random UUID in the `_authenticate()` function in `tests/function/support` so I can't just add a `testing` fxa_uid easily. 

I'm not really inclined to add a "set this user as migrating" type of call, even for testing, because that could be abused should it ever hit production. 

Any other test basically boils down to what's this test is doing, since the decorator is attached to all the relevant endpoints. 

I'll see if there's a way to get at least one full test, though.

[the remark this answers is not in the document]
jrconlin: Nevermind, I'm dumb and didn't read the patch below.

syncstorage/views/decorators.py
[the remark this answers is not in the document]
jrconlin: Yeah, totally fine with changing the comment. The code is already weirdly complex because of the fact that it's loosely bound between two packages. QA might be a bit upset that they have to wait.

syncstorage/tests/test_sql.py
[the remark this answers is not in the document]
rfk: lol; in this case  meant something a bit simpler though, basically taking the "assert that `is_migrating` ends up returning True" check from the bottom of this test, and copying it up here to the top as an "assert that `is_migrating` started out returning False" check.

syncstorage/storage/sql/__init__.py
```diff
@@ -397,6 +397,24 @@ def _find_items(self, session, user, collection, **params):
...
+        try:
+            res = session.query_fetchone("MIGRATION_CHECK", {
+                "fxa_uid": userid
+            })
+            return res is not None
+        except Exception:
+            # if the db check fails, presume the user is not migrating.
```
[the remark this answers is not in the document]
jrconlin: Adding some additional logic and passing the failure back as a 500

syncstorage/storage/sql/__init__.py
```diff
@@ -397,6 +397,25 @@ def _find_items(self, session, user, collection, **params):
             "next_offset": next_offset,
         }
 
+    @with_session
+    def is_migrating(self, session, user):
+        """Returns if the user is migrating.
```
rfk: nit: either "Returns True if" or "Returns whether"

syncstorage/storage/sql/__init__.py
```diff
@@ -397,6 +397,25 @@ def _find_items(self, session, user, collection, **params):
...
+
+        This should probably NOT be a cached call, since migration
```
rfk: In retrospect, it's not clear what you mean here by "be a cached call". Do you mean that it shouldn't be cached internally by this class, or not cached by callers of this class, or something else?

syncstorage/storage/sql/dbconnect.py
```diff
@@ -102,6 +102,22 @@ def __new__(cls, *args, **kwds):
 )
 
 
+# Table listing migration state for users
+#
+# This table contains the users in progress for migration. This table may
+# be updated by the external user migration script
+# A UID appearing in this table should cause the server to return a 5xx
```
rfk: Nit: since most instances of "uid" or "userid" in this codebase refer to the numeric id allocated by tokenserver, it may be worth mentioning explicitly that this is the fxa uid in the comment here (which you do also mention by explicit name in the table declaration below 👍)

syncstorage/storage/sql/dbconnect.py
```diff
@@ -363,6 +379,12 @@ def __init__(self, sqluri, create_tables=False, pool_size=100,
             collections.create(self.engine, checkfirst=True)
             user_collections.create(self.engine, checkfirst=True)
             batch_uploads.create(self.engine, checkfirst=True)
+            migration.create(self.engine, checkfirst=True)
+            # add this row for testing:
```
rfk: It seems unusual to have the production code add values specifically for testing, why do we need to do this here rather than e.g. in the test setup code?

syncstorage/tests/functional/support.py
```diff
@@ -45,17 +45,17 @@ def new_do_request(req, *args, **kwds):
         orig_do_request = self.app.do_request
         self.app.do_request = new_do_request
 
-    def _authenticate(self):
+    def _authenticate(self, fxa_uid=None):
         # For basic testing, use a random uid and sign our own tokens.
-        # Subclasses might like to override this and use a live tokenserver.
+        # Subclasses might like to override this and use a live tokenserver.'
```
rfk: nit: spurious trailing `'`?

syncstorage/tests/functional/test_old_storage.py
```diff
@@ -1,611 +0,0 @@
-# This Source Code Form is subject to the terms of the Mozilla Public
```
rfk: Was this file deleted on purpose? Removing it seems unrelated to the gist of the PR, but as always I could easily be missing something.

syncstorage/storage/sql/dbconnect.py
```diff
@@ -102,6 +102,22 @@ def __new__(cls, *args, **kwds):
...
+# be updated by the external user migration script
+# A UID appearing in this table should cause the server to return a 5xx
+# response. It's important that the server NEVER return a 2xx response.
+
+migration = Table(
+    "migration",
+    metadata,
+    Column("fxa_uid", String(255), primary_key=True, nullable=False),
+    Column("started_at", BigInteger, nullable=False),
+    Column("state", String(32))  # unknown/NULL, in_progress, complete
```
rfk: @Micheletto, any concern about the space overhead of using a string here rather than e.g. a smallint enum?

syncstorage/tests/test_wsgiapp.py
```diff
@@ -148,6 +139,31 @@ def new_do_request(req, *args, **kwds):
...
+        app = self._make_test_app(user)
+        req = self._make_signed_req(user["uid"], user)
+
+        # There's no public API for marking a user as migrated.
+        storage = get_storage(req)
+        with storage.dbconnector.connect() as connection:
+            connection.execute("""
+                INSERT INTO migration (fxa_uid, started_at)
+                VALUES (:fxa_uid, :started_at)
+                /* queryName=migrate */
```
rfk: nit: may want to add a `state` value here for completeness (I guess the test works because it defaults to `NULL` so it's not strictly necessary, but you use the string "in_progress" for a different test above).

syncstorage/storage/sql/dbconnect.py
rfk: (Actually, having gotten through the rest of the PR, I don't see where this value is used in any tests - maybe it's a leftover from manual testing?)

syncstorage/storage/sql/__init__.py
jrconlin: Honestly? Either. The problem right now is that migration is independent of the server, so a migration may begin during an active sync session. Since I don't believe that the migration server has the ability to alter memcache as well as the database, I don't believe that the server can properly cache this state. (I would be thrilled to be proven wrong here.)

I'm also not going to disagree that having the server check the table on call isn't fantastic. The alternate is possibly letting some corruption leak into the user record. I'm already trying my best to limit the potential by making the potential damage radius as small as possible. 

I guess it comes down to keeping the server safe or performant during migration.

syncstorage/storage/sql/dbconnect.py
jrconlin: Yeah, it's left over from a previous test that was pulled. Thanks!

syncstorage/tests/functional/test_old_storage.py
jrconlin: No, it was by accident. I had renamed the file to aid in testing. It should be restored in the latest push.

syncstorage/storage/sql/dbconnect.py
jrconlin: FWIW, I'm fine with whatever. I just need to make sure that this is coordinated with the other PRs related to migration.

jrconlin:
Not going to argue about the potential performance impact. I'm really hoping that the migration state table is VERY temporary and we can disable it once migration has occurred. If you want, I *might* be able to put it behind an option flag since most stand-alone folk are probably not going to be migrating their data to spanner anytime soon.

jrconlin:
waiting on @Micheletto's response regarding the state field.

syncstorage/storage/sql/__init__.py
rfk: Right. Just to be clear, I'm not trying to argue for any sort of caching mechanism here - it just wasn't clear as a reader of the docstring exactly what it was trying to tell me to do or not do. I would support a more direct imperative like "Do not cache the result of this call, since...".

rfk:
> If you want, I might be able to put it behind an option flag since most stand-alone folk
> are probably not going to be migrating their data to spanner anytime soon.

Self-hosters are probably least likely to be impacted by perf issues here, since they have so few users. It may be worth an option for Mozilla China though.

jrconlin:
Agreed about China. Let me see what I can do about that.

syncstorage/storage/sql/dbconnect.py
```diff
@@ -363,6 +379,7 @@ def __init__(self, sqluri, create_tables=False, pool_size=100,
             collections.create(self.engine, checkfirst=True)
             user_collections.create(self.engine, checkfirst=True)
             batch_uploads.create(self.engine, checkfirst=True)
+            migration.create(self.engine, checkfirst=True)
```
pjenvey: @Micheletto How have we done migrations (if any) in the past? Is `create_tables = true` (triggering this block) always specified in the config file?

jrconlin: From what I understand, migrations aren't always triggered, which is why i trap around any call to the migration table. The table can also be built by the migration script, which we will have a bit more control over.

jrconlin:
Going to land this because @Micheletto is going to be out for a while. Can address any concerns of his later.

syncstorage/storage/sql/dbconnect.py
erkolson: Bumping this

tublitzed: @jrconlin do we need to open a new PR here(and issue) to get this change in since it looks like this merged already?

jrconlin: Yes. see https://github.com/mozilla-services/syncstorage-rs/issues/428 &  https://github.com/mozilla-services/server-syncstorage/issues/143

PR #137, opened 2020-01-24
Add an HTTP-level test that migration produces a 503.

rfk:
I wanted to test something about the behaviour of #136, so figured I may as well do it as an actual testcase. @jrconlin r?

### DIFF
--- a/syncstorage/tests/test_wsgiapp.py
+++ b/syncstorage/tests/test_wsgiapp.py
@@ -36,14 +36,17 @@ class TestWSGIApp(StorageTestCase):
         sqluri = get_storage(req).sqluri
         self.assertTrue(sqluri.startswith("sqlite:////tmp/another-test-host-"))
 
-    def _make_test_app(self):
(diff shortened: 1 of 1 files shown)


syncstorage/tests/test_wsgiapp.py
```diff
@@ -109,20 +112,7 @@ def test_logging_of_invalid_bsos(self):
...
-        auth_policy = self.config.registry.getUtility(IAuthenticationPolicy)
-        req = Request.blank("http://localhost/")
-        auth_token, auth_secret = auth_policy.encode_hawk_id(req, user_id)
-
-        def new_do_request(req, *args, **kwds):
-            hawkauthlib.sign_request(req, auth_token, auth_secret)
-            return orig_do_request(req, *args, **kwds)
-
-        orig_do_request = app.do_request
-        app.do_request = new_do_request
```
rfk: As far as I can tell this was just duplicated code from the `make_test_app` method, so I cleaned it up while I was in here.

PR #139, opened 2020-02-05
feat: improve spanner backend compatibility

### LINKED ISSUE #135: Improve functional test_storage.py Spanner compatibility
At least one of the e2e tests run unreliably against syncstorage-rs's Spanner backend.

- test_batch_ttl_is_based_on_commit_timestamp: tests ttls w/ time.sleep and a very narrow range. It should probably give Spanner a little more leeway

pjenvey:
Closes #135

### DIFF
--- a/syncstorage/tests/functional/test_storage.py
+++ b/syncstorage/tests/functional/test_storage.py
@@ -25,6 +25,7 @@ import string
 import urllib
 import webtest
 import contextlib
+import uuid
 # import math
 
 from pyramid.interfaces import IAuthenticationPolicy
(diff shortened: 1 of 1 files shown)


syncstorage/tests/functional/test_storage.py
```diff
@@ -1964,6 +1965,15 @@ def test_that_we_dont_resurrect_committed_batches(self):
...
             batch1 = resp.json['batch']
+            try:
+                uuid.UUID(batch1)
+            except ValueError:
+                pass
+            else:
+                # On the Spanner backend: batch ids are uuid4s which
+                # should almost never collide. Plus this test will run
+                # too long, potentially expiring auth tokens
+                raise unittest2.SkipTest('N/A to Spanner backend')
```
jrconlin: I'm a bit confused on this. We check to see if `batch1` is a valid UUID, ignore if it isn't, and skip the test if there's any other problem with `batch1`?

Seems like kind of a weird fail no-op. 

Maybe a bit of info around what sort of "other" error you're skipping might help.

rfk: I interpreted this as "Skip the test if `batch` is a valid uuid, because this indicates the spanner backend, and the test provides no value for spanner"; if accurate then it may be worth refactoring the comment to indicate that at the start of the `try` block rather than inline.

PR #150, opened 2020-07-30
Fix Hawk token content validation for new fxa_kid

enote-kane:
As of mozilla-services/tokenserver#143, the Tokenserver includes the full FxA key-ID into field `fxa_kid` for the Hawk auth token, with an example value being the following:
```
1596115912183-FSK_yn0nhZYlC9CrDYZ6xg
```
...whereas it previously was just the hex-encoded client-state value, which would have been:
```
1522bfca7d278596250bd0ab0d867ac6
```
The actual issue is here, that the storage-server tries to validate the value space for `fxa_kid` after validating the token, but does not account for the value-space of url-safe Base64.

### DIFF
--- a/syncstorage/views/authentication.py
+++ b/syncstorage/views/authentication.py
@@ -20,7 +20,7 @@ DEFAULT_EXPIRED_TOKEN_TIMEOUT = 60 * 60 * 2  # 2 hours, in seconds
 # Coarse validation of FxA userid, device ids, and key ids.
 # This is not supposed to catch all invalid cases, but to act as a backstop
 # that the ids are safe to use and store internally.
-VALID_FXA_ID_REGEX = re.compile("^[A-Za-z0-9=-]{1,64}$")
+VALID_FXA_ID_REGEX = re.compile("^[A-Za-z0-9=\\-_]{1,64}$")
 
 
 class SyncStorageAuthenticationPolicy(TokenServerAuthenticationPolicy):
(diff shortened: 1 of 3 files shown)


syncstorage/views/authentication.py
```diff
@@ -20,7 +20,7 @@
 # Coarse validation of FxA userid, device ids, and key ids.
 # This is not supposed to catch all invalid cases, but to act as a backstop
 # that the ids are safe to use and store internally.
-VALID_FXA_ID_REGEX = re.compile("^[A-Za-z0-9=-]{1,64}$")
+VALID_FXA_ID_REGEX = re.compile("^[A-Za-z0-9=-_]{1,64}$")
```
rfk, line 23: I believe the `=-_` sequence there ends up doing a range match for any character between `=` and `-`. This matches some things we don't want (like `@`) and also happens to *not* match the `-` character. Could you please try adjusting this to `=_-` and see if that resolves the CI failures?

rfk: (You could also escape the `-` to `\-`, which might make it more obvious what's happening here)

rfk:
@enote-kane to clarify, does this show up for you as a `401 Unauthorized` error and an inability to sync? I'm curious why we haven't seen any issues from this in production since it seems like it should affect us in prod just as much as it would affect a self-hosted setup.

rfk:
> I'm curious why we haven't seen any issues from this in production since it seems like it should affect us in
> prod just as much as it would affect a self-hosted setup.

Looks like this is because those extra checks are not in the version currently deployed to production

syncstorage/views/authentication.py
enote-kane: Sorry for the sloppiness here. Yes, a valid regex should be `=\-_` here, otherwise it's a range. Fixed.

enote-kane:
@rfk Yes, I am getting `401 Unauthorized` for any call, even a `/info/configuration`, which effectively results in a DoS for all users.

rfk:
The CI failures here seem to be unrelated to your change, so I'm going to merge it.

PR #153, opened 2020-09-08
Use `requirements.txt` as source of truth for requirements in setup.py

rfk:
Previously, this repo had two sets of requirements: one in `requirements.txt`
that we'd use when building the project via `make`, and one in `setup.py`
that we'd use if installing it as a depenency via pip.

This has been causing problems for self-hosters, so let's have `setup.py`
just read the list of requirements out of `requirements.txt` and avoid
needing to keep two lists of requirements in sync.

(The logic here is copied directly from the tokenserver `setup.py`,
and I've taken the opportunity to copy over some other `Makefile` tweaks
while I'm here).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,9 @@
-SYSTEMPYTHON = `which python2 python | head -n 1`
+SYSTEMPYTHON = `which python2 pypy python | head -n 1`
 VIRTUALENV = virtualenv --python=$(SYSTEMPYTHON)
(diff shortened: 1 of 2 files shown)


sbraz:
@rfk I guess once this is merged, https://github.com/mozilla-services/syncserver/commit/69b89fdaa050c956172135905b72e168dcb26e41 can be reverted since `server-syncstorage` will pull it.

jrconlin (approved):

Issue #157, opened 2020-12-07
Fix failing builds (upgrade urllib3 to >= 1.25.9)


tublitzed:
Looks like we've got [build failures here](https://travis-ci.org/github/mozilla-services/server-syncstorage/builds/747697876) that are going to block future PRs from merging.

PR #158, opened 2020-12-07
task: upgrade urllib3 to fix safetycheck and dependent lib requests

tublitzed:
## Description

Noticed when I went to open a separate PR that we're failing safety checks so builds are failing. Upgraded related deps.

## Testing

Ran `make safeycheck` locally - it passes. Now let's wait for CI...

## Issue(s)

Closes [#157](https://github.com/mozilla-services/server-syncstorage/issues/157).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -42,7 +42,7 @@ pyramid==1.9.1
 pyramid-hawkauth==0.1.0
 pytz==2019.1
 repoze.lru==0.6
-requests==2.20.0
+requests==2.25.0
 rsa==4.5
 simplejson==3.8.0
 six==1.10.0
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -53,7 +53,7 @@ traceback2==1.4.0
 translationstring==1.3
 umemcache==1.6.3
 unittest2==1.1.0
-urllib3==1.24.2
+urllib3==1.25.9
```
tublitzed, line 56: This is the one that failed the check (CVE-2020-26137) - `requests` had to be bumped up at the same time.

jrconlin (approved):

PR #167, opened 2022-07-26
chore: pin safety to its last version supporting python2

pjenvey:
Marking as a draft until confirming this fixes CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,7 +48,8 @@ jobs:
           name: test
           command: |
             make test MOZSVC_SQLURI=pymysql://test:test@127.0.0.1/sync_test
-            make safetycheck
+            # issues168: currently disabled (failing due to a few vulnerabilities)
+            #make safetycheck
   test_postgres:
     docker:
       - image: debian:stable-slim
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ test:
 	./local/bin/gunicorn --paste ./syncstorage/tests/tests.ini --workers 1 --worker-class mozsvc.gunicorn_worker.MozSvcGeventWorker & SERVER_PID=$$! ; sleep 2 ; ./local/bin/python syncstorage/tests/functional/test_storage.py http://localhost:5000 ; kill $$SERVER_PID
 
 safetycheck:
-	$(INSTALL) safety
+	$(INSTALL) safety==1.8.7
 	# Check for any dependencies with security issues.
 	# We ignore a known issue with gevent, because we can't update to it yet.
 	./local/bin/safety check --full-report --ignore 25837


ethowitz:
Just tested it locally, and the safety check itself failed for a few vulnerabilities

pjenvey:
I'll just disable it for now, here's the failures reported on CI:

```
./local/bin/safety check --full-report --ignore 25837
safety report
checked 90 packages, using default DB
---
-> waitress, installed 1.4.3, affected <2.1.1, id 46436
Waitress 2.1.1 includes a fix for CVE-2022-24761: When using Waitress versions 2.1.0 and prior behind a proxy that does not properly validate the incoming HTTP request matches the RFC7230 standard, Waitress and the frontend proxy may disagree on where one request starts and where it ends. This would allow requests to be smuggled via the front-end proxy to waitress and later behavior. There are two classes of vulnerability that may lead to request smuggling that are addressed by this advisory: The use of Python's 'int()' to parse strings into integers, leading to '+10' to be parsed as '10'', or '0x01' to be parsed as '1', where as the standard specifies that the string should contain only digits or hex digits; and Waitress does not support chunk extensions, however it was discarding them without validating that they did not contain illegal characters. A workaround is available. When deploying a proxy in front of waitress, turning on functionality to make sure that the request matches the RFC7230 standard. Certain proxy servers may not have this functionality.
https://github.com/Pylons/waitress/security/advisories/GHSA-4f7p-27jc-3c36
--
-> urllib3, installed 1.25.9, affected <1.26.5, id 43975
Urllib3 1.26.5 includes a fix for CVE-2021-33503: An issue was discovered in urllib3 before 1.26.5. When provided with a URL containing many @ characters in the authority component, the authority regular expression exhibits catastrophic backtracking, causing a denial of service if a URL were passed as a parameter or redirected to via an HTTP redirect.
https://github.com/advisories/GHSA-q2q7-5pp4-w6pg
--
-> pipenv, installed 2021.5.29, affected >=2018.10.9,<=2021.11.23, id 44492
Pipenv 2022.1.8 includes a fix for CVE-2022-21668: Starting with version 2018.10.9 and prior to version 2022.1.8, a flaw in pipenv's parsing of requirements files allows an attacker to insert a specially crafted string inside a comment anywhere within a requirements.txt file, which will cause victims who use pipenv to install the requirements file to download dependencies from a package index server controlled by the attacker. By embedding malicious code in packages served from their malicious index server, the attacker can trigger arbitrary remote code execution (RCE) on the victims' systems. If an attacker is able to hide a malicious '--index-url' option in a requirements file that a victim installs with pipenv, the attacker can embed arbitrary malicious code in packages served from their malicious index server that will be executed on the victim's host during installation (remote code execution/RCE). When pip installs from a source distribution, any code in the setup.py is executed by the install process.
https://github.com/pypa/pipenv/security/advisories/GHSA-qc9x-gjcv-465w
--
-> gunicorn, installed 19.9.0, affected <19.10.0, id 40104
Gunicorn 20.0.1 fixes chunked encoding support to prevent http request smuggling attacks.
https://github.com/benoitc/gunicorn/issues/2176
https://github.com/p4k03n4t0r/http-request-smuggling#request-smuggling-using-mitmproxy-and-gunicorn
--
-> click, installed 7.1.2, affected <8.0.0, id 47833
Click 8.0.0 uses 'mkstemp()' instead of the deprecated & insecure 'mktemp()'.
https://github.com/pallets/click/issues/1752
--
-> pip, installed 20.1.1, affected <21.1, id 40291
Pip 21.1 updates its dependency 'urllib3' to v1.26.4 due to security issues.
--
-> pip, installed 20.1.1, affected <21.1, id 42559
A flaw was found in python-pip in the way it handled Unicode separators in git references. A remote attacker could possibly use this issue to install a different revision on a repository. The highest threat from this vulnerability is to data integrity. This is fixed in python-pip version 21.1.
--
make: *** [Makefile:48: safetycheck] Error 255

Exited with code exit status 2
```

jrconlin (approved):

ethowitz (approved):

